PR #17, opened 2024-01-18
Add support for Sonata PCB

marnovandermaas:
This commit assumes revision 0.2 of the board. It adds a pin file, a top level module, a clock generator and the appropriate changes to the core file. It also updates the getting started guide to use the Sonata PCB by default.

This commit is derived from a PR in the demo system: https://github.com/lowRISC/ibex-demo-system/pull/86

### DIFF
--- a/data/pins_sonata.xdc
+++ b/data/pins_sonata.xdc
@@ -1,0 +1,78 @@
+## Copyright lowRISC contributors.
+## Licensed under the Apache License, Version 2.0, see LICENSE for details.
+## SPDX-License-Identifier: Apache-2.0
+
+# Using the names in the PCB design, they should match this file with a case-insensitive search:
+# https://github.com/newaetech/sonata-pcb/tree/main
+
+## Clocks
+create_clock -period 40.000 -name mainClk -waveform {0.000 20.000} [get_ports mainClk]
+
+set_property -dict { PACKAGE_PIN P15 IOSTANDARD LVCMOS33 } [get_ports mainClk];
+
+## Reset
+set_property -dict { PACKAGE_PIN R11 IOSTANDARD LVCMOS33 } [get_ports {nrst}];
+
+## General purpose LEDs
+set_property -dict { PACKAGE_PIN B13 IOSTANDARD LVCMOS33 } [get_ports {usrLed[0]}];
+set_property -dict { PACKAGE_PIN B14 IOSTANDARD LVCMOS33 } [get_ports {usrLed[1]}];
+set_property -dict { PACKAGE_PIN C12 IOSTANDARD LVCMOS33 } [get_ports {usrLed[2]}];
+set_property -dict { PACKAGE_PIN B12 IOSTANDARD LVCMOS33 } [get_ports {usrLed[3]}];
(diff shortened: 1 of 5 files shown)


doc/guide/getting-started.md
```diff
@@ -172,9 +172,13 @@ Then run this command:
 fusesoc --cores-root=. run --target=synth --setup --build lowrisc:sonata:system
 ```
 
+For the Arty A7 use the `synth_artya7` target.
+
```
alees24, line 176: I think at this early stage it's worth making it clearer why the software image must be built before building the FPGA bitstream, please.

data/pins_sonata.xdc
```diff
@@ -0,0 +1,85 @@
...
+
+## General purpose LEDs
+set_property -dict { PACKAGE_PIN B13 IOSTANDARD LVCMOS33 } [get_ports {led_user[0]}];
+set_property -dict { PACKAGE_PIN B14 IOSTANDARD LVCMOS33 } [get_ports {led_user[1]}];
+set_property -dict { PACKAGE_PIN C12 IOSTANDARD LVCMOS33 } [get_ports {led_user[2]}];
+set_property -dict { PACKAGE_PIN B12 IOSTANDARD LVCMOS33 } [get_ports {led_user[3]}];
+set_property -dict { PACKAGE_PIN B11 IOSTANDARD LVCMOS33 } [get_ports {led_user[4]}];
+set_property -dict { PACKAGE_PIN A11 IOSTANDARD LVCMOS33 } [get_ports {led_user[5]}];
+set_property -dict { PACKAGE_PIN F13 IOSTANDARD LVCMOS33 } [get_ports {led_user[6]}];
+set_property -dict { PACKAGE_PIN F14 IOSTANDARD LVCMOS33 } [get_ports {led_user[7]}];
```
alees24, line 21: Could I ask that port names at the FPGA-specific top level be consistent with the signal names on the schematics, please? I recognise that the use of capitals may be aesthetically offensive to some, and I'd say that preserving case doesn't matter since most search functions are case-insensitive.

When someone is new to a project, the extra couple of levels of indirection and renaming can make it hard for them to pair up the RTL with the circuit schematics; they need to know that the '.xdc' file exists, and where it lives, and then to map RTL port name to FPGA pin, and FPGA pin on the schematic to signal name/device on the schematic by searching for the pin name.

Right now with simple I/O it is less of an issue. As the design grows, the more complex devices are added, and the connectivity is increased, I think it's helpful for everything to be consistent.

This presently concerns the 'usr_jtag_...' signals, 'usrled..' signals, 'usrsw..' signals and 'navsw_..' signals (as named on the schematics).

alees24 (approved):
Thanks.

HU90m (approved):
This looks good and worked great for me

doc/guide/getting-started.md
marnovandermaas: I do so a few lines before:
https://github.com/marnovandermaas/sonata-system/blob/1946ed619761084da0168b88ffe0f8ce95f31d13/doc/guide/getting-started.md?plain=1#L169

marnovandermaas:
In the last force push, I renamed the signals in the XDC to match the PCB specification as suggested by @alees24. I also removed the JTAG signals as these are currently not used in the Sonata system while the debug module is missing.

PR #20, opened 2024-02-21
CHERI baremetal

marnovandermaas:
The most important change in this PR is enabling CHERI permissions checking in Ibex, adding capability tag bits to the TLUL user fields and compiling a sanity check program.

In more detail this PR:
- Vendors in the latest CHERIoT ibex with the changes necessary for to pass Verilator's lint.
- Patches upstream TLUL to add capability tags to the user field.
- Enable CHERI protection mode in Ibex.
- Add an SRAM block for capability tag bits.
- Add a CHERI baremetal sanity check program that blinks all the LEDs on the Sonata board (inverting based on inputs).


rtl/system/sonata_system.sv
```diff
@@ -12,8 +12,8 @@
 // - Debug module.
 // - SPI for driving LCD screen.
 module sonata_system #(
-  parameter int unsigned GpiWidth = 8,
-  parameter int unsigned GpoWidth = 16,
+  parameter int unsigned GpiWidth = 13,
+  parameter int unsigned GpoWidth = 24,
```
HU90m, line 16: Is this change necessary?

HU90m (approved):

rtl/system/sonata_system.sv
marnovandermaas: Not absolutely necessary, but they are better defaults for the Sonata board. The original numbers were based on the GPIO on the Arty A7 board.

HU90m: Ah fair doozies

sw/cpp/cheri_sanity/.gitignore
```diff
@@ -0,0 +1,8 @@
+boot.S.o
+boot.cc.o
+boot.elf
+boot.dump
+boot.bin
+cpu0_irom.vhx
+cpu0_irom.vmem
+end.S.o
```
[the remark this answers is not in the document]
marnovandermaas: Thanks, that's definitely better. I've made this change now.

PR #21, opened 2024-02-26
Reset changes

marnovandermaas:
These are mostly ported over from the demo system:
https://github.com/lowRISC/ibex-demo-system/pull/96

### DIFF
--- a/sonata.core
+++ b/sonata.core
@@ -16,6 +16,7 @@ filesets:
     files:
       - rtl/fpga/top_sonata.sv
       - rtl/fpga/clkgen_sonata.sv
+      - rtl/fpga/rst_ctrl.sv
     file_type: systemVerilogSource
 
   files_artya7:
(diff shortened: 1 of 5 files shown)


sonata.core
```diff
@@ -16,6 +16,7 @@ filesets:
     files:
       - rtl/fpga/top_sonata.sv
       - rtl/fpga/clkgen_sonata.sv
+      - rtl/fpga/rst_ctrl.sv
```
HU90m, line 19: Is this chappy checked in?

HU90m (changes requested):

sonata.core
marnovandermaas: Thanks for noticing. I've just pushed it to the branch.

marnovandermaas:
> Just tested and it works on my machine and board.
> 
> I feel like I want the FPGA system and Verilator system to be as similar as possible, so would like to put `rst_ctrl` into the `sonata_system`, but that's a future decision.

Great I'll create an issue to track this.

PR #23, opened 2024-02-27
Align the `cheri_sanity` binary

HU90m:


sw/cpp/cheri_sanity/irom.ldscript
```diff
@@ -8,5 +8,6 @@ SECTIONS
 		boot.S.o(.text);
 		boot.cc.o(.text .text.* .rodata .rodata.* .data.rel.ro);
 		end.S.o(.text);
+		. = ALIGN (4);
```
nbdd0121, line 11:
```suggestion
		. = ALIGN(4);
```

marnovandermaas (approved):
Tested and works for me, thanks for the fix.

sw/cpp/cheri_sanity/irom.ldscript
HU90m: Sorry missed this, fixed with a force push. THE POWA!!!

PR #25, opened 2024-02-27
Changed pc_and_gpo.gtkw to use relative paths

HU90m:


marnovandermaas (approved):
Thanks for correcting this error.

PR #26, opened 2024-03-12
Add tracing support to the sonata system

HU90m:
Note to use tracing one needs to add ` --verilator_options='+define+RVFI'` to the build command, e.g.

```sh
fusesoc --cores-root=. \
  run --target=sim --tool=verilator --setup \
  --build lowrisc:sonata:system \
  --verilator_options='+define+RVFI'
```

### DIFF
--- a/vendor/lowrisc_ibex/ibex_top_tracing.core
+++ b/vendor/lowrisc_ibex/ibex_top_tracing.core
@@ -10,7 +10,7 @@ filesets:
       - lowrisc:ibex:ibex_top
       - lowrisc:ibex:ibex_tracer
     files:
-      - rtl/ibex_top_tracing.sv
+      - rtl/ibexc_top_tracing.sv
     file_type: systemVerilogSource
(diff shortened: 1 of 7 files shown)


vendor/lowrisc_ibex/ibex_top_tracing.core
marnovandermaas, line 1: It's better to apply these patches by actually running the vendor script:
`util/vendor.py --commit --update --verbose vendor/lowrisc_ibex.vendor.hjson`

marnovandermaas (commented):
Just change the applying the patch through the vendor script and this PR is good to go.

I tested it using `./build/lowrisc_sonata_system_0/sim-verilator/Vsonata_system -t --meminit=ram,./sw/cpp/cheri_sanity/boot.elf` and it populated the trace log file :+1:

marnovandermaas (approved):
Thanks, looks good to me!

PR #31, opened 2024-03-22
Use fork instead of patch files

marnovandermaas:
Use the CHERIoT Ibex fork instead of patch files so that it is easier to rebase changes.

This commit also updates the GTKWave file after the UART changes and bumps the CHERIoT Ibex to the latest changes.

The first vendor commit simply changes over from Microsoft's repository to our fork (so there should be no changes except the lock file). The second vendor commit is after I rebased our changes onto the latest upstream commit.


HU90m (approved):
Looks good

Verilated system builds and runs

PR #34, opened 2024-03-25
Reproducible sim build fix

HU90m:
I hadn't realised but I broke the reproducible sim build in an earlier PR


marnovandermaas (approved):
I haven't tested this but looks good to me.

PR #38, opened 2024-03-29
Fix capability tag corruption of SRAM

nbdd0121:
The address out from tlul_adapter_sram has already the last two bit trimmed. We further strip out the bits before feeding it to the cap ram, causes multiple words to be mapped to the same address in the cap ram.

I move the SRAM related part to a dedicated module so logically related parts are grouped together.

### DIFF
--- a/rtl/system/sram.sv
+++ b/rtl/system/sram.sv
@@ -1,0 +1,172 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module sram #(
+  // (Byte-addressable) address width of the SRAM.
+  parameter int unsigned AddrWidth = 17,
+  parameter InitFile               = ""
+) (
+    input  logic clk_i,
+    input  logic rst_ni,
+
+    input  tlul_pkg::tl_h2d_t tl_a_i,
+    output tlul_pkg::tl_d2h_t tl_a_o,
+
+    input  tlul_pkg::tl_h2d_t tl_b_i,
+    output tlul_pkg::tl_d2h_t tl_b_o
+  );
+
+  localparam int unsigned BusAddrWidth  = 32;
+  localparam int unsigned BusByteEnable = 4;
+  localparam int unsigned BusDataWidth  = 32;
+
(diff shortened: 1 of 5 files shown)


HU90m (approved):
Looks great

Tested in Verilator and on FPGA

rtl/system/sram.sv
```diff
@@ -0,0 +1,172 @@
...
+  logic                     mem_a_req   ;
+  logic [BusAddrWidth-1:0]  mem_a_addr  ;
+  logic                     mem_a_we    ;
+  logic [BusByteEnable-1:0] mem_a_be    ;
+  logic [BusDataWidth-1:0]  mem_a_wdata ;
+  logic                     mem_a_wcap  ;
+  logic                     mem_a_rvalid;
+  logic [BusDataWidth-1:0]  mem_a_rdata ;
+  logic                     mem_a_rcap  ;
+  logic                     mem_a_err   ;
```
marnovandermaas, line 33: Nit: We don't usually align the semicolons, it's also not done in the rest of this file.

rtl/system/sram.sv
```diff
@@ -0,0 +1,172 @@
...
+  logic                    unused_mem_b_rcap;
+
+  assign mem_a_err  = 1'b0;
+
+  // TL-UL device adapters
+
+  tlul_adapter_sram #(
+    .SramAw           ( AddrWidth ),
+    .EnableRspIntgGen ( 1         )
+  ) sram_b_device_adapter (
```
marnovandermaas, line 48: Nit: I would have preferred to put the B adapter after the A adapter.

marnovandermaas (commented):
Thanks for fixing this bug in the memory system (and apologies for introducing it)! Thanks also for moving the SRAM logic to its own file (that's been on my todo list for a while). I added some small comments as I went through.

rtl/system/sram.sv
nbdd0121: What I did in this PR is simply moving the logic from the top level to a separate file, and remove the data/inst naming and replace them with a/b. I didn't change any of the styling or order to keep the PR simple.

marnovandermaas: Thanks, that makes sense!

PR #40, opened 2024-04-03
[rtl] Add new SPI IP

GregAC:
~~For visibility only, this isn't yet ready for review/merge~~

~~I've only tested transmit so far, I've got a slightly hacky integration with ibex-demo-system that demonstrates this SPI block working with the LCD up to 25 MHz.~~

Outdated, see below.


GregAC:
Ok this is now good to go. I've tested it against the flash and demonstrated successful write and read back @ 25 MHz. Have yet to try it against the LCD with this iteration though did try the earlier version of the spi_core with a hacked up bus adapter with Ibex demo system and that worked with the LCD.

Note that interrupts are specified but not yet implemented.

I do have some example code though we need a sensible setup for it to work with so it can be included in the repository. Locally I disabled CHERI and just ran it as plain rv32e code. I'm thinking of creating some 'CHERIoT bare' software stub that will jump into a C main function with broad capabilities available so you can write little test programs without worrying about the RTOS or capabilities.

GregAC:
>  I'm thinking of creating some 'CHERIoT bare' software stub that will jump into a C main function with broad capabilities available so you can write little test programs without worrying about the RTOS or capabilities.

Ah looks like we've already got more or less what I was thinking of: https://github.com/lowRISC/sonata-system/tree/main/sw/cpp/cheri_sanity

I'll do an SPI flash test using that framework.

rtl/ip/spi/rtl/spi_core.sv
```diff
@@ -0,0 +1,289 @@
...
+      sample_phase_d = cpha_i;
+      clk_counter_d  = '0;
+    end
+  end
+
+  always_ff @(posedge clk_i) begin
+    clk_q          <= clk_d;
+    clk_counter_q  <= clk_counter_d;
+    sample_phase_q <= sample_phase_d;
+  end
```
marnovandermaas, line 183: Should we set these to default values on reset/

rtl/fpga/top_sonata.sv
marnovandermaas, line 1: We need to make sure top_verilator compiles as well (maybe just tie off these spi signals for now.

rtl/ip/spi/dv/verilator/spi_core_tb.sv
```diff
@@ -0,0 +1,187 @@
...
+
+    .cpol_i     (1'b0),
+    .cpha_i     (1'b1),
+    .msb_first_i(1'b1),
+
+    .spi_copi_o(spi_copi),
+    .spi_cipo_i(spi_cipo),
+    .spi_clk_o (spi_clk),
+
+    .half_clk_period_i(10'd5)
```
marnovandermaas, line 174:
```suggestion
    .half_clk_period_i(16'd5)
```

This solves a Verilator lint warnign.

rtl/ip/spi/rtl/spi_core.sv
GregAC: I don't think there's any need to, they'll get set to values based upon the config registers on the first clock edge out of reset anyway.

Plus they only have relevance when communicating with an SPI device which by default will be unselected and will be set to sensible values by the time the SPI block wants to do something.

GregAC:
Unfortunately the SPI flash test program isn't yet working on the board though looked ok in sim. Might be something to do with the program load via debug, I haven't tried it as initial SRAM contents.

My original C code that I ported is here: https://gist.github.com/GregAC/b4bdd5ec5d325f96cef2764c0d47b001 (that's intended to run on vanilla Ibex). That was working fine.

rtl/ip/spi/rtl/spi_core.sv
marnovandermaas: Ok, marking this as resolved.

rtl/ip/spi/dv/verilator/spi_core_tb.sv
marnovandermaas: This has been changed.

rtl/ip/spi/dv/verilator/README.md
```diff
@@ -0,0 +1,14 @@
+This is a basic testbench to exercise spi_core.
+It does not aim to fully verify the design.
+
+To build the testbench a v5 Version of verilator is required (it was testerd with v5.020).
+
+```sh
+# Build the simulation
+verilator --binary --trace-fst spi_recv.sv spi_trans.sv ../../rtl/spi_core.sv ./spi_core_tb.sv
```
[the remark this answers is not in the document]
marnovandermaas: This has been resolved.

rtl/fpga/top_sonata.sv
marnovandermaas: This has since been resolved.

marnovandermaas:
I ran the test program in CHERI mode and got the JEDEC as `0xef4019` which corresponds to the data sheet table in section 8.1.1: https://www.mouser.co.uk/datasheet/2/949/w25q256jv_spi_revg_08032017-1489574.pdf for part W25Q256JVEIQ. However the flash contents read out is all zeros.

marnovandermaas:
I've updated the LCD demo in the non-CHERI world. It turns on the display but currently does not write anything to it. Still investigating why this is.

marnovandermaas:
I found the issue, which was that `appspi_cs` was spelt as `appsci_cs` in `top_sonata.sv` as it was passed in through `gp_o`. Now the non-CHERI flash test passes in FPGA.

marnovandermaas (approved):
After my changes this now passes the SPI flash test in both CHERI and non-CHERI land.

PR #42, opened 2024-04-05
Minor source code changes

marnovandermaas:
Some minor changes that I noticed while working on integrating the debug module.

### DIFF
--- a/dv/verilator/top_verilator.sv
+++ b/dv/verilator/top_verilator.sv
@@ -13,7 +13,7 @@ module top_verilator (input logic clk_i, rst_ni);
   // Instantiating the Sonata System.
   sonata_system u_sonata_system (
     // Clock and Reset
-    .clk_sys_i (clk_i),
+    .clk_sys_i (clk_i ),
     .rst_sys_ni(rst_ni),
 
     // UART TX and RX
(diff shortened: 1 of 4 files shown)


dv/verilator/top_verilator.sv
```diff
@@ -21,27 +21,27 @@ module top_verilator (input logic clk_i, rst_ni);
...
-    .pwm_o     ( ),
-    .spi_rx_i  (0),
-    .spi_tx_o  ( ),
-    .spi_sck_o ( ),
+    .gp_i     (0),
+    .gp_o     (),
+    .pwm_o    (),
+    .spi_rx_i (0),
+    .spi_tx_o (),
+    .spi_sck_o(),
```
HU90m, line 29: I don't think I prefer this change. Is there anything about this in the style guide?

HU90m (approved):
Passed Verilator lints. Looks good bar the comment

dv/verilator/top_verilator.sv
marnovandermaas: It's just that I've formatted the rest of the RTL this way. I happy either way but then we would have to change the formatting of all the other files.

nbdd0121: I agree with Hugo. The visual alignment of parenthesis looks better.

marnovandermaas: Ok, let's slowly change the repository to follow that style then.

PR #44, opened 2024-04-10
Add rv_plic

nbdd0121:


marnovandermaas (commented):
Thanks for doing this work! How have you tested this setup? It would be nice to see that the timer interrupt we have still works as expected.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+rtl/bus/tl_main.pkg.sv linguist-generated=true
```
marnovandermaas, line 1: should be main_pkg

rtl/system/sonata_system.sv
```diff
@@ -756,6 +849,27 @@ module sonata_system #(
...
   );
 
+  tlul_pkg::tl_h2d_t tl_rv_plic_h2d_fixup;
+  always_comb begin
+    tl_rv_plic_h2d_fixup = tl_rv_plic_h2d;
+    // The address map we selected means that addr[26] is 1.
+    // The non-standard extension of OpenTitan rv_plic however extends the address space to make use of this additional bit.
+    // As a hack clear this bit.
+    tl_rv_plic_h2d_fixup.a_address[26] = 1'b0;
+  end
```
marnovandermaas, line 859: I prefer moving the PLIC to `0x8800_0000` to avoid this hack.

sw/c/common/sonata_system_regs.h
```diff
@@ -18,6 +18,8 @@
 #define I2C0_BASE 0x82000000
 #define I2C1_BASE 0x82001000
 
+#define RV_PLIC_BASE 0x84000000
```
marnovandermaas, line 21: Just a note, this should change to 88 as well.

sw/c/demo/i2c_hat_id/i2c_hat_id.c
```diff
@@ -21,6 +30,13 @@ int main(void) {
   static uint8_t data[0x80u];
 
   uart_init(DEFAULT_UART);
+  rv_plic_init();
+
+  // Test IRQ
+  rv_plic_register_irq(fmt_threshold_irq, i2c_irq_test);
+  rv_plic_enable(fmt_threshold_irq);
+  DEV_WRITE(DEFAULT_I2C + 0x4,  0x1);
+  DEV_WRITE(DEFAULT_I2C + 0x24, 0x10000);
```
marnovandermaas, line 39: Might be nice to put a few comments in here to say that this enables interrupt and set format threshold.

.gitattributes
marnovandermaas: now fixed

Issue #16, opened 2024-01-18
Remove FIFO from LSU TileLink


marnovandermaas:
This issue tracks a TODO that was specified in a commit message:

> This commit adds a `tlul_fifo_sync` to latch `tl_ibex_lsu_h2d`, this
needs to be investigated further and resolved for performance reasons.

https://github.com/lowRISC/sonata-system/pull/14/commits/562865cbbac16ef50c32ee5341a2203efa3db75c

marnovandermaas:
This is done now.

PR #46, opened 2024-04-17
Add debug module

marnovandermaas:
This is a pretty big pull request, but it was quite complex to put together so I would rather not peal it apart too much. Apologies if this makes reviewing more difficult.

You can of course do a code review, but please also test this PR. You can do so with the following commands:
```sh
pushd sw/cpp/cheri_sanity
make CHERIOT_LLVM_ROOT=<path-to>/llvm-cheriot/bin CHERIOT_RTOS_SDK=<path-to>/cheriot-rtos/sdk
popd
pushd sw/cpp/error_leds
make CHERIOT_LLVM_ROOT=<path-to>/llvm-cheriot/bin CHERIOT_RTOS_SDK=<path-to>/cheriot-rtos/sdk
popd
fusesoc --cores-root=. run --target=synth --setup --build lowrisc:sonata:system
# Please connect up your Sonata board before the following command.
openFPGALoader -c ft4232 build/lowrisc_ibex_demo_system_0/synth_sonata-vivado/lowrisc_ibex_demo_system_0.bit
# You can observe that the user LEDs start blinking and respond to joystick presses after the previous command.
./util/load_new_software_on_fpga.sh ./sw/cpp/cheri_sanity/boot.elf
# You can now observe that the capability exception LEDs start cycling through.
# Also you can press the reset button (SW5) to clear remaining user LED state.
```

The first four commits are just lint cleanups I found along the way, which you can ignore during the review process.


rtl/system/sonata_system.sv
```diff
@@ -260,15 +300,57 @@ module sonata_system #(
...
+    .tl_h_i(tl_sram_h2d_d),
+    .tl_h_o(tl_sram_d2h_d),
+    .tl_d_o(tl_sram_h2d_q),
+    .tl_d_i(tl_sram_d2h_q),
+
+    .spare_req_i(1'b0),
+    .spare_req_o(),
+    .spare_rsp_i(1'b0),
+    .spare_rsp_o()
+  );
```
HU90m, line 343: It will be nice to know where the feedback is coming from.

rtl/fpga/clkgen_sonata.sv
```diff
@@ -26,9 +26,9 @@ module clkgen_sonata (
     .COMPENSATION         ("ZHOLD"),
     .STARTUP_WAIT         ("FALSE"),
     .DIVCLK_DIVIDE        (1),
-    .CLKFBOUT_MULT        (34),
+    .CLKFBOUT_MULT        (33),
     .CLKFBOUT_PHASE       (0.000),
-    .CLKOUT0_DIVIDE       (17),
+    .CLKOUT0_DIVIDE       (33),
```
HU90m, line 31: :cry:

rtl/system/sonata_system.sv
```diff
@@ -482,7 +482,7 @@ module sonata_system #(
 
   assign cheri_en = 1'b1;
   assign cheri_en_o = cheri_en;
-  assign rst_core_n = rst_sys_ni & ~ndmreset_req;
+  assign rst_core_n = rst_sys_ni & ~ndmreset_req & ~host_req[DbgHost];
```
HU90m, line 485: :nerd_face: :smiling_imp:

HU90m (approved):
This is hacky, but it works. Nice hacking

rtl/system/sonata_system.sv
marnovandermaas: I already have an issue tracking this todo: https://github.com/lowRISC/sonata-system/issues/16

marnovandermaas:
The last force push squashes the changes to the load software script and the TCL because they are very related.

rtl/fpga/clkgen_sonata.sv
marnovandermaas: I opened up an issue to track this.

rtl/system/sonata_system.sv
marnovandermaas: Part of the same issue for now: https://github.com/lowRISC/sonata-system/issues/50

dv/verilator/top_verilator.sv
```diff
@@ -30,7 +30,14 @@ module top_verilator (input logic clk_i, rst_ni);
...
     // CHERI output
     .cheri_err_o(),
-    .cheri_en_o ()
+    .cheri_en_o (),
+
+    // User JTAG
+    .tck_i  (),
+    .tms_i  (),
+    .trst_ni(rst_ni),
+    .td_i   (),
```
GregAC, line 39: Would be worth tying these all to 0

pulp_riscv_dbg.core
```diff
@@ -21,22 +20,25 @@ filesets:
       - vendor/pulp_riscv_dbg/src/dmi_jtag.sv
     file_type: systemVerilogSource
 
-  files_vivado:
+  files_xilinx_bscane_tap:
```
GregAC, line 23: We don't want the bscane tap in Sonata at all so we should remove this.

rtl/system/jtag_id_pkg.sv
```diff
@@ -0,0 +1,18 @@
...
+
+package jtag_id_pkg;
+
+  // lowRISC JEDEC Manufacturer ID, bank 13 0xEF
+  localparam logic [10:0] JEDEC_MANUFACTURER_ID = {4'd12, 7'b110_1111};
+  localparam logic [3:0] JTAG_VERSION = 4'h1;
+
+  localparam logic [31:0] RV_DM_JTAG_IDCODE = {
+    JTAG_VERSION,          // Version
+    {12'h100,4'h1},        // Part Number
```
GregAC, line 13: This is the Ibex demo system part number, we might want to introduce a Sonata part number?

rtl/system/sonata_system.sv
```diff
@@ -254,15 +300,57 @@ module sonata_system #(
     .spare_rsp_o()
   );
 
+  // This latch is necessary to avoid circular logic. This shows up as an `UNOPTFLAT` warning in Verilator.
```
GregAC, line 303: I think the crossbar generator can add this for you automatically, using the `pipeline' option (currently set to 'false').

rtl/system/sonata_system.sv
GregAC: For resets they should come directly out of a flop to avoid any issues with glitches.

GregAC (commented):
Looks like this got merged before I had a chance to review, testing it now but not yet done.

It is odd adding a debug module has killed our clock frequency. A strong indicator of something weird happening that should really be investigated. I will try to take a look myself and see what's going wrong.

GregAC:
Well works fine with the Olimex which is good. Maybe we're messing up the FTDI config somehow? I did just copy/paste the `ftdi_layout_init` from elsewhere without working out what it actually did, other than noting it seemed to work at the time!

GregAC:
Oh and one more question, what's happening when the new program is being written to memory? There's nothing explicit to halt the CPU core so presumably it's running as we rewrite the instruction memory and then issue a reset? That could be pretty unreliable, especially if one binary allocates something as data it's actively reading and writing to and the new binary allocates the same space as code.

HU90m:
> Oh and one more question, what's happening when the new program is being written to memory? There's nothing explicit to halt the CPU core so presumably it's running as we rewrite the instruction memory and then issue a reset? That could be pretty unreliable, especially if one binary allocates something as data it's actively reading and writing to and the new binary allocates the same space as code.

Yep, the CPU core is reset with every word written from JTAG at the moment :nauseated_face: :

https://github.com/lowRISC/sonata-system/blob/a1345240eb7c872d59c7a1aeb226c600be4130cf/rtl/system/sonata_system.sv#L510

But we will probably be fine, because the CPU will likely never get past the first few lines of boot/setup-code before the next write.

pulp_riscv_dbg.core
marnovandermaas: This has now been removed: https://github.com/lowRISC/sonata-system/pull/56

rtl/system/sonata_system.sv
marnovandermaas: I opened up an issue to track this.

marnovandermaas: https://github.com/lowRISC/sonata-system/issues/66

rtl/system/jtag_id_pkg.sv
marnovandermaas: I proposed a new part number here: https://github.com/lowRISC/part-number-registry/pull/3

dv/verilator/top_verilator.sv
marnovandermaas: I've created a PR for this: https://github.com/lowRISC/sonata-system/pull/68

rtl/system/jtag_id_pkg.sv
marnovandermaas: Created an issue to track this: https://github.com/lowRISC/sonata-system/issues/69

PR #47, opened 2024-04-23
Integration of I2C block

alees24:
This PR integrates the I2C block from the OpenTitan project, and updates the software accordingly:

- RTL changes for I2C integration into FPGA and Verilator simulation
- Vendor in I2C block from OT project (please check; vendoring and patch file creation is unfamiliar)
- Modify existing software base (sw/c/common) to use the OpenTitan UART for output
- Introduce simplified I2C driver software for the OT I2C block; this is based on the code from the OT project but is rewritten to be simpler, more compact and have no dependencies such as 'base', and 'lib/dif' etc.
- Create a small test application 'i2c_hat_id' that reads and reports the contents of the ID EEPROM present on standard-compliant RaspberryPi HATs.

Please note that this demo will not run properly in Verilator simulation because there is presently no I2C DPI model to act as the target device, but may be run on Sonata FPGA by tying the 'cheri_en' low in sonata_system.sv. This code is intended not just to test the operation of the integrated block but also to serve as a guide for CHERIoT driver/application development.

PR #48, opened 2024-04-23
Integration of usbdev block

alees24:
This PR builds on #47, adding the USB device from the OpenTitan project in a similar manner to the integration of the I2C block.

- Instantiate usbdev with vendoring and patching (again, please check; this is unfamiliar)
- Clock generation is modified to provide a 48MHz clock alongside the - now parameterized - system clock; it is likely that the system clock will be changed at some point, and the initial hope was that the Ibex core could run faster than the present 50MHz.
- usbdev requires a 48MHz clock (this is 4 x oversampling of a 12Mbps signal)
- **usbdev does not have an assigned base** so I've just slotted this into the address window of the I2C devices presently; each of these IP blocks requires only a 4KiB window individually.
- Simplified code for responding to the configuration messages/set up that the USB host performs is included in sw/c/common/usbdev.c/h. This should suffice for most USB device implementations, but it is deliberately kept quite minimal; derived from the OT code, and intended as 'proof of life'/basis for further software development.
- A demo application called 'hello_usb' that mimics the behavior of 'hello_world' is introduced; this will print a sign-on message over the UART and then send a message 'Hello USB! <cycles>  Input value: <x>' over the a 'simpleserial' USB connection to the USB host, reporting the state of the switches as per 'hello_world.'

The demo application will start up in Verilator simulation but will never receive the configuration Control Transfers at present because there is no integration of the USBDPI model. With Sonata FPGA, once the 'cheri_en' signal is tied low and the appropriate software image specified (sw/c/build/demo/hello_usb/hello_usb.vmem), the application will run on FPGA and the USB output can be observed using 'screen /dev/ttyUSB<n>' where the particular tty assigned may be ascertained from the reports of USB activity in 'dmesg' output.

Any required changes to the I2C vendoring/patching (or anything else) I shall mirror here; please review that first.


marnovandermaas (approved):
I rebased this PR and tested it. It looks good to me.

PR #51, opened 2024-04-26
Introduce CHERI enable/disable switch

alees24:
CHERI shall be enabled by default. To disable CHERI functionality experimentally, eg. to run the legacy code base, set DIP switch 8 to ON before bringing the system out of reset.
If wishing to toggle a FPGA image with downloaded code between CHERI and non-CHERI, the Reset switch on the Sonata board may be used, provided that the software has not destructively modified RAM contents eg. code.


marnovandermaas (approved):
Looks reasonable to me. Will be useful as we do development.

PR #53, opened 2024-05-01
uartdpi: Ignore write failures

HU90m:
Currently if there is failure to write to the TTY or file, an assert fails and the simulation core dumps.

This patch changes that to log the error that was observed and continue the simulation.


nbdd0121:
In what circumstance would the write fail? If the TTY is not open on the other end it should just block rather than fail?

HU90m:
These are non blocking writes, so it fails with errno `EAGAIN`. We could make them blocking, but I don't want the whole sim to stop just because the TTY hasn't been connected.

marnovandermaas (approved):
This looks fine to me. I have not tested it, but if you are happy with how much you've tested this, please feel free to merge it!

PR #55, opened 2024-05-01
Impement navigation menu for the LCD demo

nbdd0121:
I need a `snprintf` implementation so I take the implementation from OpenTitan runtime.


nbdd0121:
@marnovandermaas can you review the "Fix SPI driver" and "Fix timing freq parameter" commits? They're needed to make the existing LCD demo functional again

marnovandermaas:
> @marnovandermaas can you review the "Fix SPI driver" and "Fix timing freq parameter" commits? They're needed to make the existing LCD demo functional again

@nbdd0121 Thanks for making these fixes! From reading them they look very reasonable to me. I would like to test the SPI flash test on FPGA before merging this PR so let me know when this is ready to be tested.

nbdd0121:
@marnovandermaas the code is now ready. Could you test the flash?

sw/legacy/common/spi.h
```diff
@@ -7,27 +7,26 @@
 
 #include "stdint.h"
 
-#define SPI_TX_REG 0
-#define SPI_STATUS_REG 4
+#define SPI_BASE 0x83000000
```
marnovandermaas, line 10: This needs to be removed. Because we've defined this in sonata system regs header file: https://github.com/lowRISC/sonata-system/blob/main/sw/legacy/common/sonata_system_regs.h#L19-L20

marnovandermaas (approved):
I've tested the CHERI version of the SPI test which works well. I also tested your LCD demo and it works on my board as well.

I have mentioned a change to spi.h, but even with that change the SPI test is broken. That can be fixed in a subsequent PR though.

sw/legacy/common/spi.h
marnovandermaas: Now resolved.

PR #56, opened 2024-05-01
Remove support for artya7

marnovandermaas:
This repository is only used for the Sonata board. Remove the artya7 target because the top was no longer compatible or supported.


marnovandermaas:
Rebase

HU90m (approved):

sonata.core
```diff
@@ -46,11 +39,6 @@ filesets:
       - data/pins_sonata.xdc
     file_type: xdc
 
-  files_constraints_artya7:
-    files:
-      - data/pins_artya7.xdc
```
nbdd0121, line 51: This file should be removed as well?

marnovandermaas: Thanks, done!

PR #59, opened 2024-05-02
Revocation memory

marnovandermaas:
This commit creates a two-ported SRAM to be used as revocation tags.

I've tested that regular CHERI code still works but not sure whether the revocation tags themselves work.


marnovandermaas:
In my latest force push I added a basic SRAM test for the revocation memory.

nbdd0121 (approved):
The RTL changes look alright. I haven't checked or tested the SW. There seems quite a bit SW duplication but that could be left to a later cleanup PR.

sw/cheri/revocation_test/boot.S
```diff
@@ -0,0 +1,51 @@
...
+    .type start,@function
+start:
+	// ca0 (first argument) contains the read-write root
+	cspecialr        ca0, mtdc
+
+	// Zero the tag memory
+	//li               a1, 0x200fe000
+	//csetaddr         ca0, ca0, a1
+	//li               a1, 0x20100000
+	//cjal             zero_memory
```
HU90m, line 16: Could these commented out sections be removed?

HU90m (approved):

sw/cheri/revocation_test/boot.S
marnovandermaas: I think so, this is copied from the CHERI sanity test and I left it in for reference. Let's do a more general cleanup in the software in a later PR.

PR #61, opened 2024-05-06
Fix rv_timer register offsets

alees24:
Transpose mtime and mtimecmp registers in rv_timer implementation to match CLINT specification.
Change legacy software accordingly.


HU90m (approved):
whoops-a-daisy

PR #62, opened 2024-05-06
Cleaned up the software build for cheri

HU90m:
Used CMake because it is already in use for legacy software


marnovandermaas (commented):
Thanks for making this much needed cleanup. I would like the build system to work with setting `CHERIOT_LLVM_ROOT` instead of using the default clang version on the system.

marnovandermaas:
My force push includes a rebase on top of the current main branch as well as some fixes to the a UART check test.

sonata.core
```diff
@@ -49,7 +49,7 @@ parameters:
   SRAMInitFile:
     datatype: str
     description: SRAM initialization file in vmem hex format
-    default: "../../../../../sw/cheri/cheri_sanity/cpu0_irom.vmem"
+    default: "../../../../../build_cheri/tests/cheri_sanity.vmem"
```
nbdd0121, line 52: The convention we use for ibex-demo-system and legacy SW is probably to put it in `sw/cheri/build`

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "sw/third_party/cheriot_rtos"]
+	path = sw/third_party/cheriot_rtos
+	url = https://github.com/microsoft/cheriot-rtos
```
nbdd0121, line 3: Please avoid submodules. Either use vendoring or perhaps explore `FetchContent` in CMake.

sw/cheri/README.md
```diff
@@ -0,0 +1,51 @@
+# CHERIoT programs for the Sonata system
+
+See the [getting started guide](../../doc/guide/getting-started.md) for how to install dependancies.
+
+## Building
+
+To build, run the following from the root of the directory.
+
+```
+cmake -G Ninja -B build_cheri -S sw/cheri
```
nbdd0121, line 10: I would avoid `-G Ninja` in the doc to avoid adding extra dependency. Most people have `make` installed but not necessary ninja.

sw/cheri/cheriot_toolchain.cmake
```diff
@@ -0,0 +1,27 @@
...
+
+set(CMAKE_ASM_COMPILER_TARGET riscv32-unknown-unknown)
+set(CMAKE_C_COMPILER_TARGET riscv32-unknown-unknown)
+set(CMAKE_CXX_COMPILER_TARGET riscv32-unknown-unknown)
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
+
+
+string(CONCAT CMAKE_CXX_FLAGS_INIT
```
nbdd0121, line 17: Add flags for C as well or disable C from the project?

marnovandermaas:
Another thing I noticed is that SPI test works before these changes but not after. This needs to be investigated before we can merge this.

sw/cheri/tests/spi_test.cc
```diff
@@ -169,12 +169,12 @@ class SpiFlash
     Capability<volatile uint32_t> gpo;
     uint32_t csn_bit;
 
-    static const uint8_t CmdReadJEDECId = 0x9f;
-    static const uint8_t CmdWriteEnable = 0x06;
-    static const uint8_t CmdSectorErase = 0x20;
-    static const uint8_t CmdReadStatusRegister1 = 0x05;
-    static const uint8_t CmdPageProgram = 0x02;
-    static const uint8_t CmdReadData = 0x03;
+    const uint8_t CmdReadJEDECId = 0x9f;
```
nbdd0121, line 172: I think this change is no longer necessary?

HU90m: It is still needed :frowning_face:

marnovandermaas (approved):
SPI test works now. The UART test is broken because of frequency but we can fix that in a later PR.

PR #65, opened 2024-05-07
Dual UART

alees24:
Implement and test a second UART.

- RTL changes mimic the first UART.
- Second UART is presently unconnected in Verilator sim, but transmission can be viewed.
- 'Hello world' demo application is modified to emit a sign-on message to both UARTs (verified in sim and on FPGA).
- Simple character echo from one UART to the other provides a very basic test of receive functionality on FPGA.


marnovandermaas:
I think we should hold off on adding a second UART because the RTOS does not have support for it.

HU90m:
> I think we should hold off on adding a second UART because the RTOS does not have support for it.

I'm not sure about support and how easily it can be used in the RTOS, but the demo doesn't need it so I think this is a good call

davidchisnall:
Adding support in the RTOS is absolutely trivial and would make the compartmentalisation demo nicer.

davidchisnall:
(The only change needed to the RTOS is to add it to the board description JSON)

sw/legacy/demo/hello_world/main.c
```diff
@@ -9,28 +9,63 @@
...
+  uart_out(uart1, ch1);
+}
+
 int main(void) {
 //  install_exception_handler(UART_IRQ_NUM, &test_uart_irq_handler);
 //  uart_enable_rx_int();
-  uart_init(DEFAULT_UART);
+  if (dual_uart) {
+    const char *signon = "hello_world demo application; UART ";
+    uart0 = UART_FROM_BASE_ADDR(UART0_BASE);
```
nbdd0121, line 44: Any reason this is a static variable rather than a macro definition in `sonata_system.h`?

marnovandermaas (approved):
I rebased and tested this PR so I think it's good to go in.

PR #70, opened 2024-05-07
Shrink tag RAM

alees24:
There is a single capability tag bit per 64 bits of data. With two bits per 64 bits a write to one half of the 64-bit capability will not invalidate the capability tag associated with the other half.

### DIFF
--- a/rtl/system/sonata_system.sv
+++ b/rtl/system/sonata_system.sv
@@ -323,7 +323,7 @@ module sonata_system #(
   wire dbg_release_core = &dbg_release_cnt;
   always_ff @(posedge clk_sys_i or negedge rst_sys_ni) begin
     if (!rst_sys_ni) begin
-      dbg_release_cnt  <= '0;
+      dbg_release_cnt  <= {21{1'b1}};
     end else if (host_req[DbgHost] | ~dbg_release_core) begin
       dbg_release_cnt  <= host_req[DbgHost] ? '0 : (dbg_release_cnt + 1);
     end
--- a/rtl/system/sram.sv
+++ b/rtl/system/sram.sv
@@ -32,7 +32,7 @@ module sram #(
   //
   // For all non-capability stores the tag bit will be cleared, marking those
   // 64 bits as not containing a valid capability.
-  localparam int unsigned TOff = SingleTagPerCap ? (3 - AOff) : 0;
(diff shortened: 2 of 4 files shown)


marnovandermaas (dismissed):
There is one tag per 32 bits in the CHERIoT safe repository. I think it is up to the core to make sure which tags match up, so I would rather not change this.

It looks like they should be aligned according to the ISA, which means the unaligned tags are just not used:
image

davidchisnall:
Architecturally, there is one tag bit per 64 bits of memory.  With CHERIoT Ibex, we didn't want to double the width of the memory bus and so we duplicate the tags in each half and and them on load.  This ensures that a 32-bit data write will invalidate the architectural tag, because it invalidates one of the microarchitectural tags that is anded together to get the final one.

The details of this are in the MICRO paper.

On our larger Flute prototype, we widened the memory bus to 65 bits and had a single microarchitectural tag.  Stores that were narrower than 64 bits used byte enables for the data and unconditionally wrote 0 to the tag.

It would probably (untested, may not work) also be possible to use a separate SRAM for tags and just store the last-written tag bit.  The only things from the core that ever write a 1 for either half will be writing it for both and will write a full capability.  All 1, 2, or 4-byte writes will always write a zero.  This is not safe if there are other things that can write to memory in between the two half-capability stores coming from the core (you would then be able to interleave an access and write an arbitrary 32-bit pattern for half of a capability), but if the core is the only thing that can write to memory (or if it has a burst mode that lets it write two 32-bit words atomically) then this may be safe.

alees24:
Thanks for the clear and detailed explanation. You make a good point about the possibility of an intervening write between the two 32-bit stores of the CSC operation; our bus implementation (TL-UL, TileLink Uncached Light) does not provide a mechanism for ensuring atomicity, although presently we do not have any other bus masters.

I have changed the implementation to be parameterisable for a single tag bit per capability, which for now buys us a reduction in the memory requirements. This bit is overwritten on every bus write operation to that 64-bit region, and returned on all reads from that 64-bit region.

If at some later point we have the potential for an intervening write to cause problems, eg. secondary CPU core or a peripheral block/controller that is not capability-aware, we can change the parameter.

marnovandermaas (approved):
I've added a simple test for capability tags in this PR that now satisfies me that this works and I'm happy to merge this. I also fixed the simulator.

PR #72, opened 2024-05-08
256KiB RAM, 30MHz system clock on A50T

alees24:
- Increase SRAM size to 256KiB by avoiding bit-level write enables.
- Increase system clock frequency to 30MHz (just misses timing at 33MHz presently).
- Resurrect legacy 'hello_word' demo application now that rv_timer is usable.

This build is timing clean on A50T and passes legacy code tests (memory_test, i2c_hat_id, hello_world, hello_usb) and hello_compartment from the CHERIoT RTOS world.

### DIFF
--- a/data/xbar_main.hjson
+++ b/data/xbar_main.hjson
@@ -36,7 +36,7 @@
       xbar:  false,
       addr_range: [{
         base_addr: "0x00100000",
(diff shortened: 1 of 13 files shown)


marnovandermaas (approved):
Tested this and it looks good to me.

PR #74, opened 2024-05-08
Extended reset assertion

alees24:
Update: Two separate commits to permit builds with and without the switch; combine them and build/merge if happy, please.

This PR introduces an extended reset assertion to hold the core in reset whilst the debugger is downloading code. ~~It is a draft at present because there is a DIP switch that selects between the new behaviour (SW7 in the OFF position) and the previous brief reset assertion (SW7 in the ON position).~~

**Apply the following change** in sdk/core/loader/debug.hh:

```
--- a/sdk/core/loader/debug.hh
+++ b/sdk/core/loader/debug.hh
@@ -393,12 +393,14 @@ namespace
                        if constexpr (Enabled)
                        {
                                // Ensure that the compiler does not reorder messages.
+/*
                                asm volatile("" ::: "memory");
                                writer.format("\x1b[35mloader\033[0m");
                                writer.format(": ");
                                writer.format(fmt, args...);
-                               writer.format("\n");
+                               writer.format("\r\n");
                                asm volatile("" ::: "memory");
+*/
                        }
                }
```

Build 01.hello_word with **'--debug-loader=y and --debug-scheduler=y'** to produce console output as normal.
This will lead to much more reproducible failures (fails most times in half a dozen) by killing the logging traffic that was acting as a braking mechanism, yielding failures akin to the following captured output ~~with SW7 in the ON position:~~ with earlier builds.

```
Unused bits in sizeAndPermissions are not zero, field contains 0x66220000
sdk/core/loader/types.h:../../sdk/core/loader/types.h:956 Invariant failure in size
Unused bits in sizeAndPermissions are not zero, field contains 0x66220000

../../sdk/core/loader/types.Scheduler: Created thread for trusted stack 0x1002a8 (v:1 0x1002a8-0x100370 l:0xc8 o:0xa p: G RWcgml -- ---)
Hello world compartment: Hello world

../../sdk/core/loader/types.h:95../../sdk/core/loader/types.h:../../sdk/core/loader/types.h:956 Invariant failure in size
 Unused bits in sizeAndPermissions are not zero, field contains 0x66220000
../../sdk/core/loader/types.h:956 Invariant failure in size

Unused bits in sizeAndPermissions are not zero, field contains 0x4e002210
```
Confirm also that you see verification errors on the download.

~~Throw SW7 to the OFF position and try again.~~ It has never been observed to fail with the extended reset, nor have there been any verification failures on the download.

(FPGA image was built on the ram256-a50t branch rather than main, but there should be no difference)

@johngt for awareness

### DIFF
--- a/rtl/system/sonata_system.sv
+++ b/rtl/system/sonata_system.sv
@@ -290,6 +290,21 @@ module sonata_system #(
   logic ndmreset_req;
   /* verilator lint_on IMPERFECTSCH */
 
+  // Hold the core in reset for a period after debug monitor accesses,
+  // in anticipation of further accesses whilst downloading completes.
+  //
+  // Release the code after 2^21 cycles, which will be ca. 20-100ms for
+  // typical system clock frequencies.
+  logic [20:0] dbg_release_cnt;
+  wire dbg_release_core = &dbg_release_cnt;
+  always_ff @(posedge clk_sys_i or negedge rst_sys_ni) begin
+    if (!rst_sys_ni) begin
+      dbg_release_cnt  <= '0;
+    end else if (host_req[DbgHost] | ~dbg_release_core) begin
+      dbg_release_cnt  <= host_req[DbgHost] ? '0 : (dbg_release_cnt + 1);
+    end
+  end
+
   // Tie-off unused error signals.
   assign device_err[Gpio] = 1'b0;
   assign device_err[Pwm]  = 1'b0;
@@ -710,7 +725,7 @@ module sonata_system #(
 
   assign cheri_en   = cheri_en_i;
   assign cheri_en_o = cheri_en;
-  assign rst_core_n = rst_sys_ni & ~ndmreset_req & ~host_req[DbgHost];
+  assign rst_core_n = rst_sys_ni & ~ndmreset_req & dbg_release_core;
 
   ibexc_top_tracing #(
     .DmHaltAddr      ( DebugStart + dm::HaltAddress[31:0]      ),


rtl/system/sonata_system.sv
```diff
@@ -697,7 +709,7 @@ module sonata_system #(
 
   assign cheri_en   = cheri_en_i;
   assign cheri_en_o = cheri_en;
-  assign rst_core_n = rst_sys_ni & ~ndmreset_req & ~host_req[DbgHost];
+  assign rst_core_n = rst_sys_ni & ~ndmreset_req & dbg_release_core;
```
nbdd0121, line 712: Holding the reset for a longer time sounds like a hack.

I think resetting based upon host_gnt rather than host_req might be sufficient to fix the issue?

alees24: No, resetting on the bus access request from the debug monitor is the hack, and extending the reset certainly does not solve the problem under all possible circumstances. We should not be resetting the core just because the debug monitor wants to access memory, because we shall surely want to perform debug accesses on the live or temporarily halted system at some point.

The problem presently is that the core is let loose to run between the memory writes that are performed in downloading the sections from the ELF image, which leaves it wreak havoc by running incomplete code, and modifying global static data, so that it does not then execute as expected after the 'final' reset.

(It may also be worth mentioning that memory writes from the debug monitor are certainly not going to be back-to-back, because the data delivery from OpenOCD is uncontrollable, the JTAG is at a slower clock than the system clock, and it's serial rather than 32-bit parallel...there will be dozens or hundreds of cycles between word writes, but it is in principle unbounded.)

marnovandermaas (approved):
This looks like what we had discussed and is good to include in the repo until we have a proper halting mechanism working. I tested it on FPGA and works on the previous working binaries so happy to merge.

PR #75, opened 2024-05-13
Moved UART check over to upstream driver

HU90m:
Also added simple LED check

### DIFF
--- a/sw/cheri/tests/uart_check.cc
+++ b/sw/cheri/tests/uart_check.cc
@@ -2,61 +2,18 @@
 #define CHERIOT_NO_NEW_DELETE
 #define CHERIOT_PLATFORM_CUSTOM_UART
 
+#include "../../common/defs.h"
 #include <cheri.hh>
 #include <platform-uart.hh>
 #include <stdint.h>
 
 using namespace CHERI;
 
-#define GPIO_VALUE  (0xFFFFFFFF)
-#define CPU_FREQ_HZ (30'000'000)
-#define BAUD_RATE   (   115'200)
-
-/**
- * OpenTitan UART
- */
-class OpenTitanUart
-{
-  typedef uint32_t RegisterType;
-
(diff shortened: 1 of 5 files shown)


sw/cheri/tests/uart_check.cc
```diff
@@ -68,19 +25,16 @@ extern "C" void rom_loader_entry(void *rwRoot)
...
 
   // Create a bounded capability to the UART
-  Capability<volatile OpenTitanUart> uart = root.cast<volatile OpenTitanUart>();
-  uart.address() = 0x80100000;
-  uart.bounds()  = 0x34;
+  Capability<volatile OpenTitanUart<>> uart = root.cast<volatile OpenTitanUart<>>();
+  uart.address() = UART_ADDRESS;
+  uart.bounds()  = UART_BOUNDS;
+
+  uart->init(BAUD_RATE);
```
marnovandermaas, line 32: How does the UART know the CPU frequency?

marnovandermaas (commented):
I haven't tested this yet.

sw/cheri/tests/uart_check.cc
HU90m: The `CPU_TIMER_HZ` is common among our bare metal software and the CHERIoT RTOS. *[Implementation here.](https://github.com/microsoft/cheriot-rtos/blob/e8dfab9d3955e4a536c4d334e1d94b891e08f2cb/sdk/include/platform/sunburst/platform-uart.hh#L79)*

marnovandermaas: Ok, so do we then still need it defined in: `sw/common/defs.h`?

HU90m: Yep we do

marnovandermaas (approved):
Ok, I tested these. The GPIO test works as expected. The UART test prints the "hello there" and echoes whatever I type. Great work!

PR #81, opened 2024-05-21
Moved SPI test over to upstream driver

HU90m:
Moving to our CHERIoT RTOS fork. The goal is to close any divergences between our fork and upstream as soon as possible.


marnovandermaas:
Checks actually passed but somehow the status on this PR was not being updated.

PR #83, opened 2024-05-22
I2C Resource Optimization

GregAC:
This removes target mode logic from the I2C block, reduces FIFO sizes and forces use of distributed RAM for FIFO implementation.

DO NOT MERGE - This has not been tested only used for implementation trials


marnovandermaas:
We don't need to add more I2C blocks since we merged Adrian's PR that includes 2 I2C buses. I would prefer to leave these optimizations until after the Hackathon unless we really need it.

PR #85, opened 2024-05-22
Fix English and ninja prompt

benlaurie:


marnovandermaas (approved):
Thanks for feeding this back upstream!

PR #87, opened 2024-05-22
Add support for other I2C buses

alees24:
This PR adds support for the other I2C buses on the Sonata board.

- I2C0 drives qwiic0, RPi HAT ID and Arduino.
- I2C1 drives qwiic1, RPi GPIO/I2C bus and mikroBUS Click.

The `i2c_hat_id` demo application is modified to read and report the SparkFun AS6212 Temperature Sensor reading as well as the IMU 'WHO_AM_I' registers from the other I2C bus on the RPi Sense HAT.

This is therefore reading from I2C1 (IMU) as well as two separate devices on the I2C0 bus. I2C devices are presently not available on mikroBUS Click and Arduino boards for testing.


marnovandermaas (approved):
This works for me, thanks!

PR #90, opened 2024-05-23
Hardwire CHERI to enable

marnovandermaas:
Removing a debug feature from the main branch.

### DIFF
--- a/rtl/fpga/top_sonata.sv
+++ b/rtl/fpga/top_sonata.sv
@@ -162,19 +162,9 @@ module top_sonata (
   wire scl1_i = scl1 & rph_g3_scl & mb6;
   wire sda1_i = sda1 & rph_g2_sda & mb5;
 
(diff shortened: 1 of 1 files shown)


HU90m (approved):

nbdd0121:
Instead of hardwiring it to disable, I wonder if we can have the rp2040 firmware communicating this information. It's still useful to have this ability in some form.

GregAC:
> Instead of hardwiring it to disable, I wonder if we can have the rp2040 firmware communicating this information. It's still useful to have this ability in some form.

I am wondering if we can put it under control of Ibex with some kind of reset controller. Then the bit could be included in the flash metadata. The bootloader would remain running in CHERIoT mode but it could load a plain rv32imc binary and then instruct Ibex to reboot in non-CHERIoT mode.

However definitely a post hackathon activity!

marnovandermaas:
Interesting, we would also have to have a non-CHERI bootloader that is automatically selected in legacy mode.

PR #98, opened 2024-05-24
Add pyproject.toml and build dedicated python env

nbdd0121:
Instead of pulling in all OpenTitan python deps, use a dedicated python environment. This is what the ibex-demo-system repo does.

To remove the extra added pip dependency in flake.nix, cheriot-ibex is updated to contain a patch that is present in ibex repo that removes the deprecated distutils and pip3 dependency.


marnovandermaas (commented):
Thanks for putting this together. Do we need this in before the Hackathon?

nbdd0121:
Last time I tried the OT Python environment doesn't build on MacOS. I can't try on MacOS now to tell if this slim version would work, but it should be easier to nudge it to a working state.

The sonata simulator builds fine with this environment. It is also more consistent with our python-requirements.txt.

HU90m (approved):
Both the simulator and sim build for me with this new environment

PR #99, opened 2024-05-24
DOC: Mention programming the RP2040 as a default step

colinoflynn:
Matches the release quickstart guide now. Eventually could make this step optional, but during active development I think makes sense to tell people to always do it.


colinoflynn:
You can see this live at https://newaetech.github.io/sonata-system/ too.

marnovandermaas (approved):

PR #100, opened 2024-05-24
Connect SPI for headers

marnovandermaas:
There are two SPI instances on the R-Pi hat, one on the Arduino shield and one on the mikroBUS click.

This PR has been built but not tested with physical headers.

The first three commits are code clean up so feel free to skip those and only review the last three commits.


GregAC (approved):
LGTM though we should add the new SPI blocks to the table in the SPI doc: https://github.com/lowRISC/sonata-system/blob/main/doc/ip/spi.md

marnovandermaas:
> LGTM though we should add the new SPI blocks to the table in the SPI doc: https://github.com/lowRISC/sonata-system/blob/main/doc/ip/spi.md

Good, point added that now.

PR #102, opened 2024-05-27
Verilator v5

GregAC:
I'd like this working for the hackathon so if you've got more major changes you'd like following review I'd appreciate it if we could just record it in an issue for now and get this merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,12 +50,17 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install dependencies
-      run: |
-        sudo apt install verilator libelf-dev
-        pip install -r python-requirements.txt
+    - name: Install Nix
+      uses: cachix/install-nix-action@v24
+      with:
+        nix_path: nixpkgs=channel:nixos-23.11
+        extra_nix_config: |
+          substituters = https://nix-cache.lowrisc.org/public/ https://cache.nixos.org/
+          trusted-public-keys = nix-cache.lowrisc.org-public-1:O6JLD0yXzaJDPiQW1meVu32JIDViuaPtGDfjlOopU7o= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
+
+    - name: Nix Format Check
+      run: nix fmt -- . --check
 
     - name: Build Verilator simulator
       run: |
-        fusesoc --cores-root=. run --target=sim --tool=verilator --setup --build lowrisc:sonata:system --verilator_options="+define+RVFI"
-
+        nix build .#sonata-simulator
--- a/flake.nix
+++ b/flake.nix
@@ -57,7 +57,7 @@
         pname = "sonata-simulator";
         src = ./.;
         buildInputs = with pkgs; [libelf zlib];
-        nativeBuildInputs = (with lrPkgs; [pythonEnv verilator_ot]);
+        nativeBuildInputs = [pkgs.verilator pythonEnv];
         buildPhase = ''
           HOME=$TMPDIR fusesoc --cores-root=. run \
             --target=sim --tool=verilator --setup \
--- a/flake.nix
+++ b/flake.nix
@@ -52,6 +52,8 @@
         };
       };
 
+      cheriotPkgs = lowrisc-nix.outputs.devShells.${system}.cheriot.nativeBuildInputs;
+
       sonata-simulator = pkgs.stdenv.mkDerivation rec {
         inherit version;
         pname = "sonata-simulator";
@@ -70,6 +72,20 @@
           cp -r build/lowrisc_sonata_system_0/sim-verilator/Vtop_verilator $out/bin/${pname}
         '';
         meta.mainProgram = pname;
+      };
+
+      sonata-sim-boot-stub = pkgs.stdenv.mkDerivation rec {
+        inherit version;
+        pname = "sonata-sim-boot-stub";
+        src = sw/cheri/sim_boot_stub/.;
+        nativeBuildInputs = cheriotPkgs;
+        buildPhase = ''
+          make
+        '';
+        installPhase = ''
+          mkdir -p $out/share/
+          cp -r sim_boot_stub $out/share
+        '';
       };
     in {
       formatter = pkgs.alejandra;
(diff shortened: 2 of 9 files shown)


HU90m:
This looks brilliant, thanks Greg!

I'll quickly change the GitHub actions to do simulator builds in the nix sandbox, so we don't have to keep updating the dependencies.

GregAC:
> This looks brilliant, thanks Greg!
> 
> I'll quickly change the GitHub actions to do simulator builds in the nix sandbox, so we don't have to keep updating the dependencies.

Thanks I've just added a commit to this PR to do just that but as you know nix better than me let's see what you propose as well!

.github/workflows/ci.yml
```diff
@@ -50,12 +50,14 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install dependencies
-      run: |
-        sudo apt install verilator libelf-dev
-        pip install -r python-requirements.txt
+    - name: Install Nix
+      uses: cachix/install-nix-action@v24
```
HU90m, line 54:
```suggestion
      uses: cachix/install-nix-action@v24
      
    - name: Nix Format Check
      run: nix fmt -- . --check
```

May as well add a quick Nix format check while your at it.

.github/workflows/ci.yml
```diff
@@ -50,12 +50,14 @@ jobs:
...
     - name: Build Verilator simulator
       run: |
-        fusesoc --cores-root=. run --target=sim --tool=verilator --setup --build lowrisc:sonata:system --verilator_options="+define+RVFI"
-
+        # TODO: Actually build with nix so this tests the sonata-simulator
+        # package build
+        nix develop . -c fusesoc --cores-root=. run --target=sim \
+          --tool=verilator --setup --build lowrisc:sonata:system \
+          --verilator_options="+define+RVFI -j $NIX_BUILD_CORES" \
+          --make_options="-j $NIX_BUILD_CORES"
```
HU90m, line 63:
```suggestion
        nix build .#sonata-simulator
```

This'll build the simulator in a sandbox

HU90m (changes requested):

GregAC:
Thanks @HU90m, any thoughts on how to make it use the lowRISC cache? It spends the first few minutes building python dependencies

HU90m:
This should work for now:

```sh
nix build --accept-flake-config .#sonata-simulator
```

It would be nice to get the `sonata-system` CI pushing to the cache and running on pomona, espacially once https://github.com/lowRISC/sonata-system/pull/98 is merged. @nbdd0121 (tagging for visibility, no immediate action required.)

sw/cheri/sim_boot_stub/makefile
nbdd0121, line 1: `Makefile` per convention

sw/cheri/sim_boot_stub/makefile
```diff
@@ -0,0 +1,9 @@
+CFLAGS=-target riscv32-unknown-unknown -mcpu=cheriot -mabi=cheriot \
```
nbdd0121, line 1: It'll be good that this uses cmake similar to other binaries that we have, but we can save that to a later PR.

flake.nix
```diff
@@ -36,23 +36,28 @@
...
-          [lrPkgs.python_ot];
+          [lrPkgs.python_ot] ++ cheriotPkgs ;
         buildPhase = ''
           HOME=$TMPDIR fusesoc --cores-root=. run \
             --target=sim --tool=verilator --setup \
             --build lowrisc:sonata:system \
             --verilator_options="+define+RVFI -j $NIX_BUILD_CORES" \
-            --make_options="-j $NIX_BUILD_CORES"
+            --make_options="-j $NIX_BUILD_CORES" &&
+            make -C sw/cheri/sim_boot_stub/
```
nbdd0121, line 56: We probably should put the sim stub building in another derivation.

.github/workflows/ci.yml
```diff
@@ -50,12 +50,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Install dependencies
-      run: |
-        sudo apt install verilator libelf-dev
-        pip install -r python-requirements.txt
+    - name: Install Nix
+      uses: cachix/install-nix-action@v24
```
nbdd0121, line 54: To enable cache globally for all nix commands, you can do something similar to https://github.com/lowRISC/lowrisc-nix/blob/34f1d472d6066a42c684416f5e9e755d25ec6270/.github/workflows/ci.yml#L52-L54 here.

flake.nix
```diff
@@ -43,8 +43,10 @@
         pname = "sonata-simulator";
         src = ./.;
         buildInputs = with pkgs; [libelf zlib];
-        nativeBuildInputs = (with pkgs; [python311Packages.pip verilator]) ++
-          [lrPkgs.python_ot] ++ cheriotPkgs ;
+        nativeBuildInputs =
```
nbdd0121, line 46: Squash this commit?

flake.nix
HU90m: I had a similar thought, but I'm happy for this to go in for now. In a follow up PR, we can transition to CMake and create derivations for the software builds.

sw/cheri/sim_boot_stub/makefile
GregAC: It uses make intentionally so you can build it in environments where you don't have cmake just to minimize dependencies required to build it

flake.nix
GregAC: I did actually do it that way initially, then switched it over to this as it felt like it's something you should distribute with the simulator

marnovandermaas (approved):
Works for me. Happy to merge this once CI passes.

flake.nix
nbdd0121: What could be done is to build it in a separate derivation, and then copy it to $out/share for the simulator.

GregAC: Turns out we need to do this to get the osx build to work. CHERIoT LLVM clashes with the system LLVM and the verilated C++ fails to build so will make that change.

.github/workflows/ci.yml
marnovandermaas: resolved, I believe.

GregAC:
@HU90m happy to approve this now?

nbdd0121 (approved):

HU90m (approved):

marnovandermaas:
You force pushed over my documentation change, I've made a follow up PR here: https://github.com/lowRISC/sonata-system/pull/105

Issue #101, opened 2024-05-26
cmake doesn't probe for CHERIOT_{LLVM_BIN,RTOS_SDK} env vars?


nwf:
Despite the instructions saying that we need to run
https://github.com/lowRISC/sonata-system/blob/0e024fd419f3cc88acd98544fcf6649b4897b9d6/doc/guide/building-examples.md?plain=1#L62-L63 the key occurrences of those strings in the `cmake` files herein are...
* https://github.com/lowRISC/sonata-system/blob/0e024fd419f3cc88acd98544fcf6649b4897b9d6/sw/cheri/cheriot_toolchain.cmake#L5 and
* https://github.com/lowRISC/sonata-system/blob/0e024fd419f3cc88acd98544fcf6649b4897b9d6/sw/cheri/CMakeLists.txt#L13

I am just a caveman and `cmake` frightens and confuses me, but [cmake's documentation](https://cmake.org/cmake/help/v3.29/variable/ENV.html) suggests that one needs to use `DEFINED ENV{...}` to probe at environment variables.  Assuming one also wants to support the use of `-D` on the command line for these parameters, I think that means something like this diff should be applied:

```
diff --git a/sw/cheri/CMakeLists.txt b/sw/cheri/CMakeLists.txt
index f46fc99..f17ab2e 100644
--- a/sw/cheri/CMakeLists.txt
+++ b/sw/cheri/CMakeLists.txt
@@ -11,8 +11,12 @@ if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
 endif()
 
 if(NOT DEFINED CHERIOT_RTOS_SDK)
-  FetchContent_Populate(CHERIOT_RTOS)
-  set(CHERIOT_RTOS_SDK "${cheriot_rtos_SOURCE_DIR}/sdk")
+  if (NOT DEFINED ENV{CHERIOT_RTOS_SDK})
+    FetchContent_Populate(CHERIOT_RTOS)
+    set(CHERIOT_RTOS_SDK "${cheriot_rtos_SOURCE_DIR}/sdk")
+  else()
+    set(CHERIOT_RTOS_SDK "$ENV{CHERIOT_RTOS_SDK}")
+  endif()
 endif()
 
 project(sonata_system_cheriot_sw LANGUAGES C CXX ASM)
diff --git a/sw/cheri/cheriot_toolchain.cmake b/sw/cheri/cheriot_toolchain.cmake
index 0c3d690..867b1b6 100644
[...]

PR #103, opened 2024-05-27
Resolve documentation pointing to wrong SDK

marnovandermaas:
In the documentation we should just recommend that people use cmake to pull in the correct version of the SDK.

### DIFF
--- a/doc/guide/building-examples.md
+++ b/doc/guide/building-examples.md
@@ -27,55 +27,37 @@ You can drag and drop this UF2 file into the `SONATA` drive to program the firmw
 
 ## Baremetal examples
 
-This is only for advanced users.
+This is **only for advanced users**.
 If you want to build the baremetal examples in the Sonata repo you can follow these instructions.
 
 ### Additional Toolchain Setup
 
-Besides the compiler, there are a few more features the example code depends on:
+Besides the compiler, there are a few more features the example code depends on.
 
 #### SRecord Tools
 
-The makefile assumes srecord tools, which you can install with:
+The build environment uses srecord tools, which you can install with:
 
 ```bash
 sudo apt install srecord
 ```
 
-#### CHERIoT RTOS SDK Installation
+#### CHERIoT LLVM
 
-You will need a copy of [CHERIoT RTOS](https://github.com/microsoft/cheriot-rtos/tree/main) for this section.
-
-On Windows you may need to set `git config --global core.symlinks true` *before* cloning the repository.
-
-Clone the repository somewhere, *not* into the root of the `sonata-system` directory, but at the same level as `sonata-system`:
-
-```sh
-cd ..
-git clone --recurse https://github.com/microsoft/cheriot-rtos.git
-```
-
-**IMPORTANT**: Set these two environmental variables. The first should point to the llvm build you did in the previous part,
-and the second should point to the `cheriot-rtos` repo you just checked out:
+**IMPORTANT**: Set this environmental variable to the llvm build you did in the previous part:
 
 ```sh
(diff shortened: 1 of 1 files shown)


marnovandermaas:
Partially fixes: https://github.com/lowRISC/sonata-system/issues/101

GregAC (approved):

doc/guide/building-examples.md
```diff
@@ -27,56 +27,39 @@ You can drag and drop this UF2 file into the `SONATA` drive to program the firmw
...
 export CHERIOT_LLVM_BIN=/path/to/cheriot-llvm/bin
-export CHERIOT_RTOS_SDK=/path/to/cheriot-rtos/sdk
 ```
 
 > WARNING: The path to `/path/to/cheriot-llvm/bin` should point to the *build* directory you created, not just the root checked out `cheriot-llvm` directory.
 > The path will look something like: `~/llvm-tools/cheriot-llvm/builds/cheriot-llvm/bin`
 
+Please make sure you unset the `CHERIOT_RTOS_SDK` environment variable as `cmake` will automatically pull in [the correct version](https://github.com/lowRISC/cheriot-rtos/tree/sonata).
+
 The following assume you have run the `source .venv/bin/activate` command in the terminal you are using, and you are
```
HU90m, line 58:
```suggestion
The following assumes you have run the `source .venv/bin/activate` command in the terminal you are using, and you are
```

doc/guide/building-examples.md
```diff
@@ -101,3 +84,4 @@ You can load software onto the FPGA over USB (JTAG) using:
 ```
 There are actually four different ways of loading the program - we normally use JTAG for development, but you can also
 program it into the serial flash device on the board. See the page [Programming the Sonata Software](../dev/sw-programming.md).
+
```
HU90m, line 87:
```suggestion
```

HU90m (changes requested):

HU90m (commented):
Also, there are a few sentences which are split over multiple lines.

nwf:
I'm not sure I approve of increasingly relying on `cmake` to pull in the RTOS in a section for "advanced users".  Advanced users can probably be told how to ensure that they fetch the suitable version (which, hopefully soon, is just mainline?) and might want to make changes to the RTOS itself, which would be difficult if it's hidden behind `cmake`.  Thoughs?

davidchisnall:
Why are you wrapping a build system that can target Sonata and comes with a dev container that just works in a load of CMake goo?

For reference, the first CHERIoT build system used CMake, but CMake was not sufficiently flexible to allow defining new kinds of targets and so we abandoned it.

GregAC:
I would note the software in this repository is intended for system bring-up/test purposes. People wanting to actually develop software for Sonata will use https://github.com/lowRISC/sonata-software or indeed just the upstream CHERIoT RTOS repository: https://github.com/microsoft/cheriot-rtos/.

The main branch of this repository and of the https://github.com/lowRISC/sonata-software are our live development repositories and as such in a state of flux right now. So in particular documentation is still being worked on along with deciding exactly what lives where etc.

We do not intend for cmake to be the primary build system for Sonata software, rather we will use the existing build flows provided by the CHERIoT RTOS upstream. It may remain in use in this repository for the limited use case of building small programs used for verification purposes and only of relevance to hardware engineers working on the RTL but this is not yet decided.

doc/guide/building-examples.md
marnovandermaas: Thanks.

doc/guide/building-examples.md
marnovandermaas: Thanks

PR #107, opened 2024-05-27
Run simulator on custom nix runner

nbdd0121:
This doesn't push to the cache but it should allow it use our CI machine's local nix store so would speed up the simulator build by quite a bit.


nbdd0121:
Reduce CI time by 4 minutes.

hcallahan-lowrisc (approved):

PR #112, opened 2024-05-28
Connect up rest of the header pins

marnovandermaas:
This includes R-Pi HAT, Arduino Shield, mikroBUS Click and PMOD.

### DIFF
--- a/data/pins_sonata.xdc
+++ b/data/pins_sonata.xdc
@@ -85,27 +85,27 @@ set_property -dict { PACKAGE_PIN K2  IOSTANDARD LVCMOS18 } [get_ports {usrusb_su
 set_property -dict { PACKAGE_PIN K1  IOSTANDARD LVCMOS18 } [get_ports {usrusb_vbusdetect}];
 
 ## PMOD0
-# set_property -dict { PACKAGE_PIN H14 IOSTANDARD LVCMOS33 } [get_ports {pmod0_1}];
-# set_property -dict { PACKAGE_PIN F16 IOSTANDARD LVCMOS33 } [get_ports {pmod0_2}];
-# set_property -dict { PACKAGE_PIN F15 IOSTANDARD LVCMOS33 } [get_ports {pmod0_3}];
-# set_property -dict { PACKAGE_PIN G14 IOSTANDARD LVCMOS33 } [get_ports {pmod0_4}];
-# set_property -dict { PACKAGE_PIN J13 IOSTANDARD LVCMOS33 } [get_ports {pmod0_5}];
-# set_property -dict { PACKAGE_PIN E17 IOSTANDARD LVCMOS33 } [get_ports {pmod0_6}];
-# set_property -dict { PACKAGE_PIN D17 IOSTANDARD LVCMOS33 } [get_ports {pmod0_7}];
-# set_property -dict { PACKAGE_PIN K13 IOSTANDARD LVCMOS33 } [get_ports {pmod0_8}];
+set_property -dict { PACKAGE_PIN H14 IOSTANDARD LVCMOS33 } [get_ports {pmod0[0]}];
+set_property -dict { PACKAGE_PIN F16 IOSTANDARD LVCMOS33 } [get_ports {pmod0[1]}];
+set_property -dict { PACKAGE_PIN F15 IOSTANDARD LVCMOS33 } [get_ports {pmod0[2]}];
+set_property -dict { PACKAGE_PIN G14 IOSTANDARD LVCMOS33 } [get_ports {pmod0[3]}];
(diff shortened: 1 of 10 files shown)


GregAC (approved):

GregAC:
Merging so I can rebase the RGB LED controller on top of it

rtl/fpga/top_sonata.sv
```diff
@@ -163,6 +206,68 @@ module top_sonata (
...
+
+  logic [15:0] ard_gp_oe;
+  logic [15:0] ard_gp_o;
+
+  logic [15:0] pmod_gp_oe;
+  logic [15:0] pmod_gp_o;
+
+  // UART input
+  assign ser0_rx_aggr = ser0_rx | mb8;
+  assign ser1_rx_aggr = ser1_rx | rph_rxd0;
```
alees24, line 222: Unless I'm mistaken, UART lines are high in the Idle state, so to combine two UART streams requires the use of AND rather than OR and you will also need to enable the internal pull ups of the FPGA for when pins are undriven.

rtl/fpga/top_sonata.sv
```diff
@@ -163,6 +206,68 @@ module top_sonata (
...
+
+  // UART output
+  assign mb7 = ser0_tx;
+  assign rph_txd0 = ser1_tx;
+
+  // R-Pi header GPIO
+  assign rph_g4  = rp_gp_oe[0]  ? rp_gp_o[0]  : 1'bZ;
+  assign rph_g5  = rp_gp_oe[1]  ? rp_gp_o[1]  : 1'bZ;
+  assign rph_g6  = rp_gp_oe[2]  ? rp_gp_o[2]  : 1'bZ;
```
alees24, line 231: Perhaps this mapping of GPI/O pins to register bits is temporary, but I am a bit concerned that gaps have not been left for the unconnected pins of the R-Pi Header and Arduino Shield; it's complicating the mapping, and creating a maintenance problem for software, especially if the remaining pins are added later.

data/pins_sonata.xdc
```diff
@@ -177,32 +197,52 @@ set_property -dict { PACKAGE_PIN M16 IOSTANDARD LVCMOS33 } [get_ports rph_g3_scl
 # set_property PULLUP true [get_ports rph_g3_scl]
 ## rev 0.5+ have on-board pull ups
 
+## UART
+set_property -dict { PACKAGE_PIN K17 IOSTANDARD LVCMOS33 } [get_ports rph_txd0];
+set_property -dict { PACKAGE_PIN L18 IOSTANDARD LVCMOS33 } [get_ports rph_rxd0];
+
+## Other GPIO
+set_property -dict { PACKAGE_PIN N15 IOSTANDARD LVCMOS33 } [get_ports rph_g4];
```
alees24, line 205: Perhaps not a priority, but it appears that a number of the R-Pi pins are not connected, particularly, eg. the SPI signals.

alees24 (commented):
Comments left, thanks.

GregAC:
@alees24 @marnovandermaas given the various things Adrian has highlighted I think we're better off reverting this for now. We're out of time to get things working for the hackathon so we'll need to do this at a later point.

data/pins_sonata.xdc
```diff
@@ -177,32 +197,52 @@ set_property -dict { PACKAGE_PIN M16 IOSTANDARD LVCMOS33 } [get_ports rph_g3_scl
...
+## GPIO
+set_property -dict { PACKAGE_PIN N17 IOSTANDARD LVCMOS33 } [get_ports ah_tmpio16];
+## SPI SCLK
 set_property -dict { PACKAGE_PIN P14 IOSTANDARD LVCMOS33 } [get_ports ah_tmpio15];
-## CIPO
+## SPI CIPO
 set_property -dict { PACKAGE_PIN R18 IOSTANDARD LVCMOS33 } [get_ports ah_tmpio14];
-## Chip select
+## SPI chip select
 set_property -dict { PACKAGE_PIN U16 IOSTANDARD LVCMOS33 } [get_ports ah_tmpio10];
```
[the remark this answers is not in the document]
marnovandermaas: Yes, I had assumed 14, 15, and 17 would be connected up to 11, 12, 13 on the Adruino side, which is also the case for the Uno. I'm happy to switch it to 11-13 though. Will do this in a follow-up PR.

rtl/fpga/top_sonata.sv
marnovandermaas: Agreed, I'll create dedicated UARTs in a follow-up PR.

data/pins_sonata.xdc
marnovandermaas: They are if you look above this piece of code: SPI1, SPI0, I2C and UART are all connected.

PR #117, opened 2024-05-30
[rtl] More thorough LED clearing on power up

GregAC:
It's been observed that the existing LED clearing on power up was not sufficient as the LEDs were turning on on FPGA reset.

This adds a time deley before we first clear the LEDs as well as a repeated clear to ensure it takes effect.

Additionally some comments have been added to the existing RTL.


GregAC:
I'd like to get this in ASAP so we can update the v0.2 release with it.

Issue #5, opened 2023-12-01
PLIC: exact interrupt mapping


marnovandermaas:
The exact mapping of sources to interrupts still need to be published.

Before we can do this we must finalize the number of interrupts each block has and how blocks that have multiple instances are mapped.

PR #120, opened 2024-06-06
Add list of interrupts to PLIC documentation

marnovandermaas:
Closes https://github.com/lowRISC/sonata-system/issues/5

### DIFF
--- a/doc/ip/plic.md
+++ b/doc/ip/plic.md
@@ -5,9 +5,35 @@ Please find more details in [the official specification](https://github.com/risc
 Part of these details are is the memory map, which shows all registers from the `base` until `base + 0x3FFFFFC`.
 Not all of these registers are mapped, most importantly we only have one core, so only one context (context 0).
 
-In terms of interrupts, the exact mapping of sources to interrupts will be published here at a later point.
-There are a number of considerations.
+The following table shows the current list of interrupts that are fed through the PLIC.
+For more detailed descriptions of what each interrupt means, please refer to the documentation of each individual hardware IP block.
+More interrupts may be added to this table in the future.
 
-- The following blocks have interrupts: timer, UART, I2C, SPI, ...
-- There are multiple UART, I2C and SPI blocks in the system.
-- Some blocks have multiple interrupts.
+| Number | Block     | Interrupt description |
+|--------|-----------|-----------------------|
+|  0     | None      | Tied to zero
+|  1,  9 | UART 0, 1 | Transmit watermark
+|  2, 10 | UART 0, 1 | Receive watermark
+|  3, 11 | UART 0, 1 | Transmit empty
+|  4, 12 | UART 0, 1 | Receive overflow
+|  5, 13 | UART 0, 1 | Receive frame error
+|  6, 14 | UART 0, 1 | Receive break error
+|  7, 15 | UART 0, 1 | Receive timeout
+|  8, 16 | UART 0, 1 | Receive parity error
+| 17, 32 | I2C 0, 1  | Format FIFO threshold
+| 18, 33 | I2C 0, 1  | Receive FIFO threshold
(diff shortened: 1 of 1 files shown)


doc/ip/plic.md
```diff
@@ -5,9 +5,35 @@ Please find more details in [the official specification](https://github.com/risc
...
+| 23, 38 | I2C (0, 1)  | SDA interference |
+| 24, 39 | I2C (0, 1)  | Stretch timeout |
+| 25, 40 | I2C (0, 1)  | SDA unstable |
+| 26, 41 | I2C (0, 1)  | Command complete |
+| 27, 42 | I2C (0, 1)  | Transmit stretch |
+| 28, 43 | I2C (0, 1)  | Transmit threshold |
+| 29, 44 | I2C (0, 1)  | Acquire FIFO full |
+| 30, 45 | I2C (0, 1)  | Unexpected stop |
+| 31, 46 | I2C (0, 1)  | Host timeout |
+| 47     | Ethernet    | Packet ready to read via the SPI connected to the ethernet chip |
```
HU90m, line 38: nit: If you're not aligning, I think it looks neater to just remove the trailing `|`.

```suggestion
| 47     | Ethernet    | Packet ready to read via the SPI connected to the ethernet chip
```

doc/ip/plic.md
```diff
@@ -5,9 +5,35 @@ Please find more details in [the official specification](https://github.com/risc
...
-In terms of interrupts, the exact mapping of sources to interrupts will be published here at a later point.
-There are a number of considerations.
+The following table shows the current list of interrupts that are fed through the PLIC.
+For more detailed descriptions of what each interrupt means, please refer to the documentation of each individual hardware IP block.
+More interrupts may be added to this table in the future.
+
+| Number | Block       | Interrupt description |
+|--------|-------------|-----------------------|
+|  0     | None        | Tied to zero |
+|  1,  9 | UART (0, 1) | Transmit watermark |
```
HU90m, line 15: nit: This is more readable for me.

```suggestion
|  1,  9 | UART 0, 1 | Transmit watermark |
```

HU90m (approved):
Thanks Marno

doc/ip/plic.md
marnovandermaas: Agreed, removed those.

doc/ip/plic.md
marnovandermaas: Agreed, removed parentheses.

doc/ip/plic.md
```diff
@@ -5,9 +5,35 @@ Please find more details in [the official specification](https://github.com/risc
...
+| 23, 38 | I2C 0, 1  | SDA interference
+| 24, 39 | I2C 0, 1  | Stretch timeout
+| 25, 40 | I2C 0, 1  | SDA unstable
+| 26, 41 | I2C 0, 1  | Command complete
+| 27, 42 | I2C 0, 1  | Transmit stretch
+| 28, 43 | I2C 0, 1  | Transmit threshold
+| 29, 44 | I2C 0, 1  | Acquire FIFO full
+| 30, 45 | I2C 0, 1  | Unexpected stop
+| 31, 46 | I2C 0, 1  | Host timeout
+| 47     | Ethernet  | Packet ready to read via the SPI connected to the ethernet chip
```
GregAC, line 38: Does the ethernet chip really only have a single interrupt source? I would assume there are multiple things that can cause an interrupt, so a more generic 'Interrupt from external SPI ethernet chip' would be a more suitable description.

marnovandermaas: So there is only one interrupt pin from the KSZ8851SNLI-TR chip, but you're right that the description should be more generic as the type of interrupt is specified in the interrupt status register.

marnovandermaas: I've made a PR with an update: https://github.com/lowRISC/sonata-system/pull/144

PR #125, opened 2024-06-17
Update lowrisc_ibex to lowrisc/cheriot-ibex@a745108a

marnovandermaas:
This pulls in the latest changes from upstream up to: https://github.com/microsoft/cheriot-ibex/tree/8fe0ec632dbb1f231f2c15b3d05008b55af46b37

I've run all the current test on FPGA and it seems to work fine. It also still compiles for simulation and the UART check works.


HU90m:
> Simulation is building via FuseSoC but not via Nix.

Nix builds via FuseSoC as well. It's likely an environment problem.

What version of Verilator are you using? (We are currently on version 5.022 in Sonata.)

marnovandermaas:
I'm using Verilator 5.024 2024-04-05 rev v5.024

marnovandermaas:
Scratch this. It was because I had a `book` directory...

HU90m (approved):
Thanks for doing this Marno

PR #127, opened 2024-06-17
Connect header GPIOs

marnovandermaas:
This is a second attempt at this PR: https://github.com/lowRISC/sonata-system/pull/112

There is some feedback on that PR which I have mostly resolved.

The second commit is just formatting, which you can skip during reviewing.


alees24 (approved):
Thanks for addressing the issues with the previous PR.

PR #131, opened 2024-06-20
Fade capability LEDs when switching off

HU90m:
This is done with a new `pwm_fade` block, placed between the error signal from the core and the led output.

### DIFF
--- a/rtl/system/pwm_fade.sv
+++ b/rtl/system/pwm_fade.sv
@@ -1,0 +1,51 @@
+// Copyright lowRISC contributors.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * This module will assert it's output, `modulated_o`,
+ * when `impulse_i` is asserted. When `impulse_i` is dropped,
+ * it will fade it's output by decreasing it's pulse width.
+ *
+ * Total fade time = (1 << PwmCounterSize)^2 * NumPeriods * the clock period
+ * The default fade time is 2.24 second, when the clock is 30MHz
+ */
+module pwm_fade #(
+  /// The size of the pwm counter.
+  parameter PwmCounterSize = 7,
+  /// The number of pwm periods between pulse width reductions.
+  parameter NumPeriods = (1 << 12) - 1
+)(
+  input logic clk_i,
+  input logic rst_ni,
+  input logic impulse_i,
+
+  output logic modulated_o
+);
+  localparam PwmCounterMax = (1 << PwmCounterSize) - 1;
+
+  logic [$clog2(NumPeriods+1)+PwmCounterSize:0] counter;
+  logic [PwmCounterSize-1:0] pwm_counter, pulse_width;
+
+  // The bottom `PwmCounterSize` bits of the counter
+  // are used as the pwm counter.
+  assign pwm_counter = counter[PwmCounterSize-1:0];
+
+  assign modulated_o = pwm_counter < pulse_width;
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin : main
+    if (!rst_ni) begin
+      pulse_width <= 0;
+      counter <= 0;
+    end else if (impulse_i != 0) begin
+      pulse_width <= PwmCounterMax;
+      counter <= 0;
+    end else if (pulse_width != 0) begin
+      if (counter == (NumPeriods << PwmCounterSize)) begin
+        pulse_width <= pulse_width - 1;
+        counter <= 0;
+      end else begin
+        counter <= counter + 1;
+      end
+    end
+  end : main
+endmodule : pwm_fade
--- a/rtl/system/sonata_system.sv
+++ b/rtl/system/sonata_system.sv
@@ -798,6 +798,17 @@ module sonata_system #(
   assign cheri_en_o = cheri_en;
   assign rst_core_n = rst_sys_ni & ~ndmreset_req & dbg_release_core;
 
+  logic [CheriErrWidth-1:0] cheri_err;
+
+  for (genvar i = 0; i < CheriErrWidth; ++i) begin : gen_pwm_fade
+    pwm_fade u_pwm_fade (
+      .clk_i       (clk_sys_i      ),
+      .rst_ni      (rst_core_n     ),
+      .impulse_i   (cheri_err[i]   ),
+      .modulated_o (cheri_err_o[i] )
+    );
+  end
+
   ibexc_top_tracing #(
     .DmHaltAddr      ( DebugStart + dm::HaltAddress[31:0]      ),
     .DmExceptionAddr ( DebugStart + dm::ExceptionAddress[31:0] ),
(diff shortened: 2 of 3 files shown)


marnovandermaas (commented):
You shouldn't need both `tick_counter` and `counter`. Can't you use the PWM module we already have in Sonata? https://github.com/lowRISC/sonata-system/blob/main/rtl/system/pwm.sv

marnovandermaas:
Ah, you need the impulse which the current module doesn't have.

HU90m:
> You shouldn't need both `tick_counter` and `counter`.

I've reused the lower bits of the `tick_counter`, for the *pwm* `counter`.

marnovandermaas:
> > You shouldn't need both `tick_counter` and `counter`.
> 
> I've reused the lower bits of the `tick_counter` for the pwm `counter`.

Thanks for making this change. I now understand better why you need the two counters. I think it may be better to use your previous design as this removes the requirement that tick counter needs to be a power of two.

marnovandermaas:
I suggest we change it like this: https://github.com/marnovandermaas/sonata-system/tree/fade-reuse-pwm

HU90m:
> Thanks for making this change. I now understand better why you need the two counters. I think it may be better to use your previous design as this removes the requirement that tick counter needs to be a power of two.

The number of ticks doesn't have to be a power of two. It just has to be larger than the maximum counter value, which is true of both designs. The only difference is that using the `pwm` module is less efficient because it doesn't use the same counter for both.

Less efficient designs are fine, if they are more readable; but I don't believe adding the `pwm` module makes this block more readable, e.g. it's harder to see you won't get a full pulse if the number of ticks is smaller than the maximum counter value. (There also isn't a check for this.)

good spot:

```sv
  logic [$clog2(NumTicks+1):0] tick_counter;
```

rtl/system/pwm_fade.sv
```diff
@@ -0,0 +1,53 @@
...
+ * This module will assert it's output, `modulated_o`,
+ * when `impulse_i` is asserted. When `impulse_i` is dropped,
+ * it will fade it's output by decreasing it's pulse width.
+ *
+ * Total fade time = (1 << CounterSize) * NumTicks * the clock period
+ * The default fade time is 1.68 second, when the clock is 30MHz
+ */
+module pwm_fade #(
+  parameter CounterSize = 7,
+  parameter NumTicks = (1 << 18) - 1
```
marnovandermaas, line 14:
```suggestion
  parameter NumTicks = (1 << 12)
```

rtl/system/pwm_fade.sv
```diff
@@ -0,0 +1,53 @@
...
+
+  output logic modulated_o
+);
+  localparam CounterMax = (1 << CounterSize) - 1;
+
+  if ($clog2(NumTicks) < CounterSize) begin : check_counter_width
+    $fatal("There are too few ticks, either lower the counter size or raise the number of ticks.");
+  end
+
+  logic [$clog2(NumTicks+1):0] tick_counter;
```
marnovandermaas, line 28:
```suggestion
  logic [$clog2(NumTicks+1)+CounterSize:0] tick_counter;
```

rtl/system/pwm_fade.sv
```diff
@@ -0,0 +1,53 @@
...
+  input logic rst_ni,
+  input logic impulse_i,
+
+  output logic modulated_o
+);
+  localparam CounterMax = (1 << CounterSize) - 1;
+
+  if ($clog2(NumTicks) < CounterSize) begin : check_counter_width
+    $fatal("There are too few ticks, either lower the counter size or raise the number of ticks.");
+  end
```
marnovandermaas, line 26:
```suggestion
```

rtl/system/pwm_fade.sv
```diff
@@ -0,0 +1,53 @@
...
+
+  always_ff @(posedge clk_i or negedge rst_ni) begin : main
+    if (!rst_ni) begin
+      pulse_width <= 0;
+      tick_counter <= 0;
+    end else if (impulse_i != 0) begin
+      pulse_width <= CounterMax;
+      tick_counter <= 0;
+    end else if (pulse_width != 0) begin
+      if (tick_counter == NumTicks) begin
```
marnovandermaas, line 45:
```suggestion
      if (tick_counter == (NumTicks << CounterMax)) begin
```

rtl/system/pwm_fade.sv
```diff
@@ -0,0 +1,53 @@
+// Copyright lowRISC contributors.
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * This module will assert it's output, `modulated_o`,
+ * when `impulse_i` is asserted. When `impulse_i` is dropped,
+ * it will fade it's output by decreasing it's pulse width.
+ *
+ * Total fade time = (1 << CounterSize) * NumTicks * the clock period
```
marnovandermaas, line 9:
```suggestion
 * Total fade time = (1 << CounterSize)^2 * NumTicks * the clock period
```

marnovandermaas (commented):
Ok, you've convinced me to leave the existing PWM module out. Then my final suggestion is to do away with the check by making number of ticks represent a number of counter max periods.

marnovandermaas:
Maybe easier, this is the diff:
```diff
diff --git a/rtl/system/pwm_fade.sv b/rtl/system/pwm_fade.sv
index 3f25d1e..b547c33 100644
--- a/rtl/system/pwm_fade.sv
+++ b/rtl/system/pwm_fade.sv
@@ -6,12 +6,12 @@
  * when `impulse_i` is asserted. When `impulse_i` is dropped,
  * it will fade it's output by decreasing it's pulse width.
  *
- * Total fade time = (1 << CounterSize) * NumTicks * the clock period
+ * Total fade time = (1 << CounterSize)^2 * NumTicks * the clock period
  * The default fade time is 1.68 second, when the clock is 30MHz
  */
 module pwm_fade #(
   parameter CounterSize = 7,
-  parameter NumTicks = (1 << 18) - 1
+  parameter NumTicks = (1 << 12)
 )(
   input logic clk_i,
   input logic rst_ni,
@@ -21,11 +21,7 @@ module pwm_fade #(
 );
   localparam CounterMax = (1 << CounterSize) - 1;
 
-  if ($clog2(NumTicks) < CounterSize) begin : check_counter_width
-    $fatal("There are too few ticks, either lower the counter size or raise the number of ticks.");
-  end
-
-  logic [$clog2(NumTicks+1):0] tick_counter;
+  logic [$clog2(NumTicks+1)+CounterSize:0] tick_counter;
   logic [CounterSize-1:0] counter, pulse_width;
 
   // The bottom `CounterSize` bits of the tick_counter
@@ -42,7 +38,7 @@ module pwm_fade #(
       pulse_width <= CounterMax;
       tick_counter <= 0;
     end else if (pulse_width != 0) begin
-      if (tick_counter == NumTicks) begin
+      if (tick_counter == (NumTicks << CounterSize)) begin
         pulse_width <= pulse_width - 1;
         tick_counter <= 0;
       end else begin
~
```

marnovandermaas (approved):
Thanks for making these changes!

PR #133, opened 2024-06-20
Revert "Update lowrisc_ibex to lowrisc/cheriot-ibex@a745108a"

marnovandermaas:
This reverts commit 7c7baf7b1f37605829edb0bd09e344273a929f11.

Bumping CHERIoT Ibex still runs all the baremetal tests and the bootloader but no longer runs CHERIoT RTOS successfully. Let's revert this commit until we investigate further.


marnovandermaas:
Created issue to track this: https://github.com/lowRISC/sonata-system/issues/134

nbdd0121 (approved):

Issue #134, opened 2024-06-20
RTOS not booting with bumped Ibex


marnovandermaas:
The RTOS no longer boots on the bumped Ibex. Investigate why this is.

Related to: https://github.com/lowRISC/sonata-system/pull/133

marnovandermaas:
The issue is the ROM loader transfers control to the loaded firmware with `cjr cra`. The `cjr cra` instruction requires that the jump target is a valid return sentry in the new version of Ibex. Can we change our bootloader to do this?

marnovandermaas:
This is the line that is affected: https://github.com/lowRISC/sonata-system/blob/514682fa066babeee25df9c1f8c42514f6d2b6c6/sw/cheri/boot/boot.S#L34

marnovandermaas:
And this change should fix it: https://github.com/CHERIoT-Platform/cheriot-safe-uart-boot-rom/blob/82fbc153075b2532b124a9eb1c223d1e61666e3b/boot.S#L39

PR #135, opened 2024-06-26
Fix bootloader to avoid return sentry

marnovandermaas:
`cjr cra` requires `cra` to be a return sentry in newer version of CHERIoT Ibex. Use `ct0` instead to avoid this problem.

Based on this commit:
https://github.com/CHERIoT-Platform/cheriot-safe-uart-boot-rom/commit/82fbc153075b2532b124a9eb1c223d1e61666e3b

Fixes: https://github.com/lowRISC/sonata-system/issues/134


HU90m (approved):
Tested, works great, thanks Marno

PR #139, opened 2024-07-05
Removed instruction tracing from default build and added simulator lint

HU90m:
Instruction tracing (using RVFI) has been removed from the default simulator build, because it is not always needed but slows the simulation and produces a large file. The new verilator lint is run against the traced version to check tracing doesn't break.


marnovandermaas (approved):
This all looks reasonable to me. I haven't tested it though.

PR #141, opened 2024-07-09
Hardware revoker

marnovandermaas:
This pull request adds the control register and interrupt logic for the hardware revoker.

The base address of the control registers is `0x8000_A000` and the interrupt is mapped at 72.

It's marked as draft until we can properly test this.


doc/ip/revoker.md
```diff
@@ -0,0 +1,15 @@
...
+It essentially sweeps memory for revoked capabilities.
+For more information on how this works please refer to [Section 3.3.3 (background pipelined revoker) of this paper](https://cheriot.org/papers/2023-micro-cheriot-uarch.pdf).
+
+These are the register offsets:
+
+| Offset | Register |
+|--------|----------|
+| 0x0000 | Start address to begin sweeping
+| 0x0004 | End address to stop sweeping
+| 0x0008 | Set when sweeping is happening (most sifnificant byte is set to constant 0x55)
```
nwf-msr, line 13:
```suggestion
| 0x0008 | Set when sweeping is happening (most significant byte is set to constant 0x55)
```

HU90m:
Passed the test suite in simulation: https://gist.github.com/HU90m/2bd03a2719f4883ee90f6eeb98ffd0e0

doc/ip/revoker.md
marnovandermaas: Thanks for spotting this!

vendor/cheriot_safe/ip/msftDvIp_mmreg.sv
```diff
@@ -0,0 +1,162 @@
...
+          6'h1: tbre_end_addr   <= reg_wdata_i;
+          6'h5: tbre_intr_en    <= reg_wdata_i[0];
+          default:;
+        endcase
+      end
+
+      if (wr_op && (reg_addr_i[7:2] == 6'h2))
+        tbre_go <= 1'b1;
+      else
+        tbre_go <= 1'b0;
```
alees24, line 84: Is this really intended behavior, that 'tbre_go' should be asserted for just a single cycle?
It looks as though that's okay for the FSM within the core, but it means that the register at offset 'h4 will not indicate the activity of the revoker.

vendor/cheriot_safe/ip/msftDvIp_mmreg.sv
```diff
@@ -0,0 +1,162 @@
...
+
+      tbre_stat_q <= tbre_stat;
+      
+      if (tbre_done)
+        tbre_epoch <= tbre_epoch + 1;
+
+      if (tbre_done) tbre_intr_stat <= 1'b1;
+      else if (tbre_intr_en) tbre_intr_stat <= 1'b0;   // auto clear
```
alees24, line 92: Similar question about the interrupt status bit; when the interrupt is enabled the status bit will not be readable via register at offset 'h10.
Worse, attempting to use the revoker without enabling the interrupt will cause a second operation to appear to complete immediately because the 'tbre_intr_stat' bit will remain asserted from the first operation, right?

alees24 (dismissed):
If I understand this correctly, I think the auto-clear logic for a couple of the status bits needs attention to make them useful and prevent a race on the 'tbre_intr_stat' bit, as annotated in the comments.

doc/ip/revoker.md
```diff
@@ -0,0 +1,15 @@
+# CHERIoT hardware revocation engine
+
+This block is integrated from the [CHERIoT Safe repository](https://github.com/microsoft/cheriot-safe/blob/main/src/msft_cheri_subsystem/msftDvIp_mmreg.sv).
+It essentially sweeps memory for revoked capabilities.
```
davidchisnall, line 4: It is the thing that does the revocation, it is looking for capabilities to freed memory.

doc/ip/revoker.md
```diff
@@ -0,0 +1,15 @@
...
+
+These are the register offsets:
+
+| Offset | Register |
+|--------|----------|
+| 0x0000 | Start address to begin sweeping
+| 0x0004 | End address to stop sweeping
+| 0x0008 | Set when sweeping is happening (most significant byte is set to constant 0x55)
+| 0x0012 | Set when sweeping is done
+| 0x0014 | Enable bit for raising an interrupt when sweeping is done
```
davidchisnall, line 15: These don't seem quite right:

0x08 is the control word.  The high bits are 0x55, writing the low bit triggers the revoker to start, reading is unspecified.

0x0c is the epoch word.  It should be a monotonic (wrapping) counter that is incremented when the revocation starts and when it completes (so the low bit is 1 during a revocation sweep).

0x10 is the interrupt status word.  We never actually use this in software.

0x12 is not a 32-bit aligned address.

0x14 is the interrupt-enable word.  This is used to request an interrupt when the current revocation sweep ends.

vendor/cheriot_safe/ip/msftDvIp_mmreg.sv
davidchisnall: We never look at the interrupt status bit in software.  The only use we would have for it is to not request an interrupt if one is already scheduled and it's cheaper to just do the store unconditionally than do load, branch, store.

vendor/cheriot_safe/ip/msftDvIp_mmreg.sv
davidchisnall: That sounds like the correct behaviour: writing 1 here starts revocation and is ignored if revocation is already running.  Reading is unspecified.

The current state is polled by the register at offset 0xc, which contains the epoch.

doc/ip/revoker.md
marnovandermaas: Thanks, I've updated this now.

doc/ip/revoker.md
marnovandermaas: Thanks, I've reworded this.

vendor/cheriot_safe/ip/msftDvIp_mmreg.sv
marnovandermaas: I agree this is a bit weird to expose as a register. Although I don't want to change the behaviour of the vendored in code. Maybe it's worth writing our own control logic in the future.

vendor/cheriot_safe/ip/msftDvIp_mmreg.sv
marnovandermaas: I've made a PR to remove it from the read interface upstream: https://github.com/microsoft/cheriot-safe/pull/17

marnovandermaas: I've actually changed this PR now to only allow it to be readable/pollable when interrupts are disabled. The original code also did not allow for clearing the status when interrupts were disabled so I fixed that too.

HU90m (approved):

PR #143, opened 2024-07-20
Increased the default baud rate

HU90m:
The motivation for changing this is to speed up the CHERIoT RTOS test suite that spends a lot of time waiting for UART.
Changing to 921,600 from 115,200 decreased the runtime from 20 to 6 minutes.

I chose 921,600 because it is the closest standard baud rate to the maximum possible baud rate of 937,500.[^1] This is very close the the maximum, but the error (0.55 cycles) hasn't caused problems in simulation or on FPGA.

[^1]: The UART divides the clock by 16 (for over sampling on the receive side) and then by 2, so the maximum baud rate is the clock divided by 32. The clock is currently 30,000,000 so the maximum baud rate is 937,500. The closest standard baud rate below this is 921,600.

### DIFF
--- a/doc/guide/building-examples.md
+++ b/doc/guide/building-examples.md
@@ -1,6 +1,7 @@
 # Building Examples
 
-The following contains some simple examples you can build for the Sonata board. Once you've got these builds working, you can easily add more features to the example code.
+The following contains some simple examples you can build for the Sonata board.
+Once you've got these builds working, you can easily add more features to the example code.
 
 ## Using our template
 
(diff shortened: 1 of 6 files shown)


nbdd0121:
Can you check that this works under Windows? I recall having issues with custom baud rate in Windows.

HU90m:
That's good to know, thanks Gary. I'll don't have a windows laptop to test this with.

GregAC (approved):
This looks good to me but as Gary said we should test in Windows.

We do have a windows laptop in the office that we can use for this.

I note this stack overflow answer: https://stackoverflow.com/a/34134778 seems to indicate the 921'600 rate should work.

HU90m:
I've checked with the echo example on windows. All seems to be in order :police_officer:

PR #149, opened 2024-08-09
Hyperram

GregAC:


GregAC:
HyperRAM is disabled by default but can be built by adding the `--EnableHyperram` flag to your fusesoc build, e.g.

```
fusesoc --cores-root=. run   --target=sim --tool=verilator --setup   --build lowrisc:sonata:system   --verilator_options="-j $NUM_CORES"   --make_options="-j $NUM_CORES" --EnableHyperram
```

For verilator or 

```
fusesoc --cores-root=. run --target=synth --setup --build lowrisc:sonata:system --SRAMInitFile=$PWD/sw/cheri/build/tests/hyperram_test.vmem --EnableHyperram
```

For FPGA build (note this particular builds in the hyperram test for the SRAM contents)

With HyperRAM enabled you loose half of the SRAM. I think Ibex may lock-up if you try to access to the top 128k of SRAM as the crossbar memory map does not get adjusted in the HyperRAM configuration.

The plan is HyperRAM will be in the default system so this problem will go away. If we do end up with a HyperRAM and non HyperRAM configs with differing SRAM sizes we can set something up so the top 128k of SRAM gives a bus error in the HyperRAM config.

I think access latency is around 10-11 cycles in this setup. Some scope to improve there but I think I've done all the easy stuff (it was 17 cycles) and we'll be getting into the heart of the controller and how it captures data coming back from the HyperRAM. Worth doing at some point but not a priority for now.

I've got a block level testbench for the HyperRAM as I completely rewrote the top-level to have a native tilelink interface. I haven't shared that as it's a bit of a mess (in particular how you build it). I do want to add them to the repository though. It works using a BFM of the HyperRAM.

I've also got a full system simulation including HyperRAM BFM you can run under Xcelium but against this requires some clean-up before it's suitable to upstream.

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+
+#include <cheri.hh>
+#include <platform-uart.hh>
+#include <stdint.h>
+#include <ds/xoroshiro.h>
+
+using namespace CHERI;
+
+const int RandTestBlockSize = 256;
+const int HyperramSize = (1024 * 1024) / 4;
```
GregAC, line 16: Note this is intended as a soak test on the real board (i.e. leave it running and see if anything fails, hopefully good for flushing out reliability issues). It has an infinite loop that runs the same tests over and over.

You can run it on Verilator but I'd suggest adjusting these numbers so it runs far fewer tests per loop. E.g. change this to:

```
const int RandTestBlockSize = 16;
const int HyperramSize = 16;
```

Once we have the baremetal testsuite in we can adapt this so we've got a soak mode and a terminating test mode.

GregAC:
Oh and this allows execution from hyperram, though it will be slow until we sort out the icache.

rtl/system/sram.sv
```diff
@@ -69,7 +69,7 @@ module sram #(
     .tl_o        (tl_a_o),
 
     // Control interface.
-    .en_ifetch_i (prim_mubi_pkg::MuBi4False),
+    .en_ifetch_i (prim_mubi_pkg::MuBi4True),
```
marnovandermaas, line 72: Why does this need to be true? The A-side of this SRAM model is connected to the LSU not the fetch stage.

marnovandermaas (commented):
Had a brief look at this and had one question. Not yet approving because I haven't had a chance to test this yet.

marnovandermaas:
Also, there's a CI linting error:
```
ERROR: Setup failed : Cannot find lint/hyperram.vlt in :
>      ./rtl/ip/hyperram
>      ./rtl/ip/hyperram
```

rtl/ip/hyperram/rtl/hbmc_tl_top.sv
```diff
@@ -0,0 +1,595 @@
...
+      hbmc_arst_sync_idelayctrl
+      (
+          .clk   ( clk_i     ),
+          .arst  ( sync_rst   ),
+          .rst   ( idelayctrl_rst )
+      );
+
+
+      (* IODELAY_GROUP = C_IODELAY_GROUP_ID *)
+      IDELAYCTRL
```
HU90m, line 164: For those who also hadn't come across this before: https://www.xilinx.com/products/intellectual-property/util_idelay_ctrl.html

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+  hyperram_area.bounds()  = HYPERRAM_BOUNDS;
+
+
+  Capability<Capability<volatile uint32_t>> hyperram_cap_area =
+    root.cast<Capability<volatile uint32_t>>();
+  hyperram_cap_area.address() = HYPERRAM_ADDRESS;
+  hyperram_cap_area.bounds()  = HYPERRAM_BOUNDS;
+
+	while (true)
+	{
```
HU90m, line 264: nit: mixing tabs and spaces. Don't worry about fixing; I'll set up an auto formatter this week.

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+  Capability<void> root{rwRoot};
+
+
+  // Create a bounded capability to the UART
+  Capability<volatile OpenTitanUart<>> uart =
+    root.cast<volatile OpenTitanUart<>>();
+  uart.address() = UART_ADDRESS;
+  uart.bounds()  = UART_BOUNDS;
+
+  uart->init(921600);
```
HU90m, line 245: nit: buad rate defined in `sw/common/defs.h`

```suggestion
  uart->init(BAUD_RATE);
```

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+  set_console_mode(uart, CC_BOLD);
+  write_str(uart, "\r\n\r\nGet hyped for hyperram!\r\n");
```
HU90m, line 247: :rofl:

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+
+    if (read_val != rand_val) {
+      failures += 1;
+    }
+  }
+
+  return failures;
+}
+
+int rand_cap_test(Capability<volatile uint32_t> hyperram_area,
```
HU90m, line 103: nit: some brief comments above these functions would be nice.

HU90m (approved):
This looks good and is working on my FPGA; I'm hyped!

I couldn't test the simulation build because of the missing `lint/hyperram.vlt`. Once the sim is building, I'm happy for this to go in. Don't worry about fixing my comments; none of them are big enough to block this going in.

GregAC:
Thanks for the reviews so far.

Just done a quick update with the missing lint file and a missing addition to the sonata verilator lint.

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+
+    Capability<volatile uint32_t> write_cap;
+    Capability<volatile uint32_t> read_cap;
+
+    do {
+      rand_addr = prng() % HyperramSize;
+      rand_cap_addr = prng() % HyperramSize;
+    } while ((rand_addr / 2) == (rand_cap_addr / 2));
+
+    rand_cap_addr /= 2;
```
nbdd0121, line 123: The intention would be clearer if written this way?
```suggestion
    do {
      rand_index = prng() % HyperramSize;
      rand_cap_index = prng() % (HyperramSize / 2); // Pointer is double word in size.
    } while (rand_index / 2 == rand_cap_index);
```

I think if a variable have name `addr`, it's probably better that it's in byte-addressed, otherwise it can be quite confusing.

nbdd0121:
When synthesising this I get 3 critical warnings:
* `[Synth 8-2490] overwriting previous definition of module ws281x_drv ["sonata-system/build/lowrisc_sonata_system_0/src/lowrisc_sonata_system_0/rtl/fpga/ws281x_drv.sv":5]`
* `[Vivado 12-4739] set_false_path:No valid object(s) found for '-from [get_pins -of [get_cells u_sonata_system/g_hyperram.u_hyperram/u_hbmc_tl_top/hbmc_ufifo_inst/u_fifo/*storage*/*]]'. ["sonata-system/build/lowrisc_sonata_system_0/src/lowrisc_sonata_system_0/data/pins_sonata.xdc":310]`
* `[Vivado 12-4739] set_false_path:No valid object(s) found for '-from [get_pins -of [get_cells u_sonata_system/g_hyperram.u_hyperram/u_hbmc_tl_top/hbmc_dfifo_inst/u_fifo/*storage*/*]]'. ["sonata-system/build/lowrisc_sonata_system_0/src/lowrisc_sonata_system_0/data/pins_sonata.xdc":311]`

GregAC:
> I think one thing that the current test misses is interleaved/randomized write to the HyperRAM with different tag bits, and read back to see if they're expected. I think if the HyperRAM just always return `1` for capability then the current test wouldn't be able to capture that?

I also have a block-level testbench that does full random read/write accesses (including different sizes, not just the 32-bit reads/writes the test here does). This uses a hyperram BFM and was seeing no failures.

It isn't included in this PR as it's a bit of a mess and isn't vital for the v1.0 release. However I do plan to make a PR for it soon.

So we can extend the C test but it's not intended to be a fully comprehensive test, more of a combination of a smoke (i.e. make sure the basics work) and a soak (i.e. keep testing things over and thing to check reliability and hopefully flush out CDC issues) test.

GregAC:
Oh and the block-level testbench does also test capability tags (randomized and modelled along with the memory data)

sw/cheri/tests/hyperram_test.cc
```diff
@@ -0,0 +1,294 @@
...
+    if (read_val != rand_val) {
+      failures += 1;
+    }
+  }
+
+  return failures;
+}
+
+int rand_cap_test(Capability<volatile uint32_t> hyperram_area,
+    Capability<Capability<volatile uint32_t>> hyperram_cap_area,
```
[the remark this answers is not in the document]
GregAC: I did that originally but you cannot have a `volatile Capability<..>` as various methods in the class need to be marked as `volatile` and that may be not be desirable for the general case and this is a pretty niche usage so manual compiler fences seemed preferable.

GregAC:
I believe I've addressed all the comments, thanks for the reviews.

Additionally as discussed (on Zulip I think) hyperram is now enabled by default, though we still have the awkward top half of SRAM missing in that configuration (because we have to define a 256k window for it to allow the no hyperram configuration).

Now to fix lint, I think I need to update the build so lint builds with the hyperram sim model (bit of a hack but verilator cannot build the full hyperram controller right now because it uses Xilinx IP, we can hopefully setup something so we can at least lint it with blackboxes for the Xilinx specific stuff but it's not an immediate priority).

GregAC:
> When synthesising this I get 3 critical warnings:
> 
>     * `[Synth 8-2490] overwriting previous definition of module ws281x_drv ["sonata-system/build/lowrisc_sonata_system_0/src/lowrisc_sonata_system_0/rtl/fpga/ws281x_drv.sv":5]`
> 
>     * `[Vivado 12-4739] set_false_path:No valid object(s) found for '-from [get_pins -of [get_cells u_sonata_system/g_hyperram.u_hyperram/u_hbmc_tl_top/hbmc_ufifo_inst/u_fifo/*storage*/*]]'. ["sonata-system/build/lowrisc_sonata_system_0/src/lowrisc_sonata_system_0/data/pins_sonata.xdc":310]`
> 
>     * `[Vivado 12-4739] set_false_path:No valid object(s) found for '-from [get_pins -of [get_cells u_sonata_system/g_hyperram.u_hyperram/u_hbmc_tl_top/hbmc_dfifo_inst/u_fifo/*storage*/*]]'. ["sonata-system/build/lowrisc_sonata_system_0/src/lowrisc_sonata_system_0/data/pins_sonata.xdc":311]`
, 
I'm not too worried about the warnings on the false path thing, I've managed to sort out the timing errors caused by false paths which is good enough for now.

As explained in the XDC ideally we'd have a more general setup for these but I've had a lot of trouble getting it to work. It seems the environment and design for the XDC execution is not the same as the environment and design used when you execute commands in the Vivado interactive TCL shell. So stuff that works just fine doesn't work when included in an XDC.

PR #151, opened 2024-08-12
Use `llvm-objcopy` to generate binaries

elliotb-lowrisc:
Need to use our LLVM version of objcopy, not a generic one that wont know how to handle the ELF output from our LLVM clang.


marnovandermaas (commented):
Can you check whether this change also works under the Nix environment?

elliotb-lowrisc:
Just tried this:
```
nix develop
export CHERIOT_LLVM_BIN=~/llvm-tools/cheriot-llvm/builds/cheriot-llvm/bin
cmake -B sw/cheri/build -S sw/cheri
cmake --build sw/cheri/build
```
Which seemed to work.

Should I have done some cleaning first or some other commands?

marnovandermaas (approved):
Looks good to me.

PR #153, opened 2024-08-15
Reset SPI flash before accessing it

nbdd0121:
If the system is reset while a SPI access is going on, the SPI device might not be in the correct state and using it directly will fail, and we don't show helpful message in this case, just "Failed ELF Magic Check".

Add a software reset sequence to reset it to a well known state before accessing.

### DIFF
--- a/sw/cheri/common/flash-utils.hh
+++ b/sw/cheri/common/flash-utils.hh
@@ -1,4 +1,5 @@
 #pragma once
+#include "timer-utils.hh"
 #include <cheri.hh>
 #include <platform-gpio.hh>
 #include <platform-spi.hh>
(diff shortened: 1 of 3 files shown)


sw/cheri/common/timer-utils.hh
```diff
@@ -0,0 +1,20 @@
...
+static inline uint32_t get_mcycle(void)
+{
+	uint32_t result;
+	__asm__ volatile("csrr %0, mcycle;" : "=r"(result));
+	return result;
+}
+
+static inline void reset_mcycle(void)
+{
+	__asm__ volatile("csrw mcycle, x0");
```
HU90m, line 13: Why not?

```suggestion
	asm volatile("csrw mcycle, x0");
```

HU90m (approved):

sw/cheri/common/flash-utils.hh
```diff
@@ -32,6 +35,20 @@ class SpiFlash
 	{
 	}
 
+	void reset()
+	{
+		set_cs(true);
+		spi->blocking_write(&CmdEnableReset, 1);
+		set_cs(false);
+
+		set_cs(true);
```
marnovandermaas, line 44: Should we wait a few cycles between these two?

Issue #148, opened 2024-07-31
Potential pinmux design


GregAC:
@alees24 @marnovandermaas please note below.

Here's an outline of a design spec for the pinxmux work.  Feel free to suggest different ways to build this, just wanted to get my view of how it would all work written down and thought it'd be useful to have some discussion of the design before we implement it.

The pinmux will allow multiple blocks within Sonata to control a single physical pin.  It does not enable dynamic any to any muxing functionality (where any IO from any block can control any pin).  Rather a physical pin will have a small number of block IOs it could connect to.  E.g. a particular pin could choose between a GPIO, CSn of SPI block 1, RX of UART 1 or SDA of I2C 1.  There will be two parts to the pinmux:

 - A static configuration, which is used to generate the pinmux RTL from a configuration file.
   - This provides full flexibility, for each physical pin it will list the possible block IO connections
 - The dynamic mux, what actually exists in RTL and is controlled by software, it chooses the block that connects to the pin, given the choices defined by the static configuration

Each physical pin will provide the following interface:

```verilog
input  pin_output_i;        // Value to output on the pin
input  pin_output_enable_i; // Pin driven to output value when set, otherwise pin_output_i is ignored
output pin_input_o;         // Input value from the pin (always valid will be equal to pin_output_i when pin_output_enable_i is set)
```

Enable programmatic access to pull-up/pull-down configurations could be considered but isn't a critical part of core functionality.

The software interface for the pinmux will consist of a number of registers each filled with copies of an n-bit field (fixed n across all registers). Each field corresponds to a particular physical pin. Writing to that field sets which block controls that pin. The meaning of the index for each field will vary from pin to pin depending on what block IO is available to connect to that physical pin. We may wish to reserve index 0 to mean unconnected (so output enable low and input routed nowhere)

When a pin is connected to a block there are three possible types:
 - output is disabled and pin input is fed to the block
 - output is enabled, pin input isn't routed anywhere and block control pin output
 - block given direct access to input, output and output enable to manipulate as it wishes

Which type is used is part of the static configuration

## Static configuration

The static configuration has two major sections

**Block Defintions** - Split into two sub-sections, block types and block instances.

A block type provides the name of a block and of the IO that block (giving a name and type for each IO). Block instances specify how many of each block type is available.

An example YAML configuration for the block definition could look like this:

```yaml
- block: spi
  ios:
    - pin: cs_n
      type: output
    - pin: sck
      type: output
    - pin: tx
      type: output
[...]

PR #154, opened 2024-08-16
Pinmux and top generation

marnovandermaas:
This implements the pinmux. It generates the number of GPIO, UART, I2C and SPI blocks specified by the `data/blocks.yml`. There is a `util/top_gen.py` script that uses templates to generate the appropriate changes to the crossbar and the package that specifies the number of each block.

This PR also does the following useful things:
- Adds output enables to the GPIO module.
- Make it so that multiple GPIO registers can live in the same 4 KiB memory.
- Reduced number of SPI blocks from 7 to 5.

There are a number of possible refinements to this design which can be addressed as follow-up PRs and I have captured them here: https://github.com/lowRISC/sonata-system/issues/169

Closes: #148


sw/legacy/common/uart.c
```diff
@@ -10,6 +10,7 @@
 #define BAUD_RATE (921600)
 
 void uart_enable_rx_int(void) {
+  // TODO this must now instead be done using the PLIC
```
GregAC, line 13: Code cleanup here isn't relevant to the rest of the commit

sw/legacy/demo/ethernet/ksz8851.c
```diff
@@ -222,7 +222,7 @@ static void ksz8851_recv(struct netif *netif) {
 
     // Reset QMU RXQ frame pointer to zero.
     ksz8851_reg_write(spi, ETH_RXFDPR, 0x4000);
-    
+
```
GregAC, line 225: Is this a tab to space change? (Or vice versa hard to tell in the github UI...)

data/blocks.yml
```diff
@@ -0,0 +1,36 @@
...
+  instances: 2
+  ios:
+    - name:    scl
+      type:    in_out
+      combine: and
+    - name:    sda
+      type:    in_out
+      combine: and
+spi:
+  instances: 7
```
GregAC, line 19: With the pinmux we should seek to reduce the number of SPI and UART instances, we've got so many as we've needed one per physical interface.

Need to think carefully about i2c as well, it's a big block so good resource savings by reducing the number. Though still if we're not using the resources for anything else may as well have more i2c, 2 feels like a reasonable number though we could likely manage with 1.

data/pins.yml
```diff
@@ -0,0 +1,437 @@
...
+      instance: 2
+      io:       rx
+ethmac_sclk:
+  block_ios:
+    - block:    spi
+      instance: 2
+      io:       sck
+
+# Raspberry Pi HAT
+rph_g0:
```
GregAC, line 103: This setup is fine for getting things going but for the v1.0 release I do want to enable more flexibility, e.g. so a given set of rpi HAT pins could be GPIO, could be I2C, could be UART or could be SPI. If you restrict yourself to what the rpi can actually do (uncertain how flexible it is in connecting its own SPI etc blocks to the GPIO pins) you may have a more restricted set but it is also a general purpose header people can connect non RPI HAT stuff to so flexibility is useful.

util/top_gen.py
```diff
@@ -4,11 +4,91 @@
 # SPDX-License-Identifier: Apache-2.0
 
 from mako.template import Template
+import yaml
 
-gpio_num = 5
-uart_num = 5
-i2c_num  = 2
-spi_num  = 7
+# First we parse the YAML input to get the parameters.
```
GregAC, line 9: We should decouple the pinmux generation from the top gen and make sure validation/parsing functionality is nicely separated.

As we've deprioritised top gen I'd focus entirely on a standalone pinmux tool that's written with reasonable modularity so at a later stage it could be used within topgen rather than run as a separate invocation.

util/top_gen.py
```diff
@@ -4,11 +4,91 @@
...
+
+with open('data/blocks.yml', 'r') as file:
+    block_parameters = yaml.safe_load(file)
+with open('data/pins.yml', 'r') as file:
+    pin_parameters = yaml.safe_load(file)
+
+# Check block parameter structure.
```
GregAC, line 16: We should use an existing YAML schema validation library, e.g. https://pypi.org/project/schema/ (I'm sure there are more there's another one I used in the past but cannot remember what it's called but schema looks pretty decent from a quick glance).

That said for now we should minimise validation work, just putting some asserts in places where you're using the data structure pulled from the YAML to check things are as you expect should do for now. We need to prioritise getting the pinmux generated and usable over good error checking etc.

GregAC (commented):
Was there a reason to do the auto-generated top? I guess it's only a modest amount of work in this PR  but we had de-prioritised the topgen/system builder work for the v1.0 release.

sw/legacy/demo/ethernet/ksz8851.c
marnovandermaas: This removes trailing whitespace.

util/top_gen.py
marnovandermaas: Using Pydantic instead. Thanks to @HU90m for the work.

sw/legacy/common/uart.c
marnovandermaas: Moved this to a cleanup commit.

sw/legacy/demo/ethernet/ksz8851.c
marnovandermaas: Moved this to a cleanup commit with message saying what's done.

data/pins.yml
marnovandermaas: I've captured this here: https://github.com/lowRISC/sonata-system/issues/169

util/top_gen.py
marnovandermaas: I've captured this here: https://github.com/lowRISC/sonata-system/issues/169

data/blocks.yml
marnovandermaas: I've captured this here: https://github.com/lowRISC/sonata-system/issues/169

rtl/system/pinmux.sv
```diff
@@ -101,23 +101,78 @@ module pinmux (
...
+  logic [BusDataWidth-1:0] reg_wdata;
+  logic [(BusDataWidth/8)-1:0] reg_be;
+  logic [BusDataWidth-1:0] reg_rdata;
+
+  logic unused_reg_signals;
+
+  //TODO allow reading selector values.
+  assign reg_rdata = BusDataWidth'('0);
+
+  tlul_adapter_reg #(
```
GregAC, line 124: I feel like it'd be better to use reggen to generate all of this. It is a pain but it keeps things consistent with the other IPs. We've already got an issue in Sonata that we've got some IPs using the legacy bus from Ibex demo system each with their own custom register logic, this adds another kind of custom register logic. Long term it's better if everything is using the same methods consistently (i.e. reggen for everything)l

So pinmux generator would also produce a hjson that can then be fed through reggen. This probably means we need to vendor reggen into this repository and patch it so it doesn't generate the integrity related stuff. As the SPI block uses reggen and currently the reggen output has been manually altered to remove the integrity stuff it is something we want anyway. Especially as we should also move the legacy IPs on to reggen to get consistency.

GregAC (commented):
I appreciate this need to be merged to avoid having to keep rebasing it and avoiding conflicts and not all comments from my review need to be addressed but I do view the use of Zs within pinmux and the combination of pinmux/pad functionality as critical things that must be addressed before we can merge.

I would like to get the reggen stuff sorted before v1.0 as well as the updated GPIO module (where we can have a single GPIO block with multiple sets of registers to control all GPIO pins that can live on a single port on the crossbar).

rtl/system/sonata_system.sv
```diff
@@ -8,10 +8,8 @@
 module sonata_system #(
   parameter int unsigned GpiWidth      = 13,
   parameter int unsigned GpoWidth      = 24,
-  parameter int unsigned RpiGpiWidth   = 11,
-  parameter int unsigned ArdGpiWidth   = 10,
-  parameter int unsigned PmodGpiWidth  = 16,
   parameter int unsigned ArdAniWidth   = 6,
+  parameter int unsigned HalfWordWidth = 16,
```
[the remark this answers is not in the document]
marnovandermaas: Ok, I have now removed this HalfWordWidth in the new changes.

data/xbar_main.hjson
```diff
@@ -299,6 +269,56 @@
         size_byte: "0x00001000",
       }],
     },
+    { name:  "gpio0", // General purpose input and output
+      type:  "device",
+      clock: "clk_sys_i",
+      reset: "rst_sys_ni",
+      xbar:  false,
+      addr_range: [{
+        base_addr: "0x80500000",
```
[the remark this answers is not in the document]
marnovandermaas: I've now removed these extra GPIO modules and implemented a gpio_array module that instantiates multiple GPIO register groups in the same address space.

data/top_config.toml
```diff
@@ -0,0 +1,392 @@
...
+  { name = "sck", type = "output" },
+  { name = "tx", type = "output" },
+  { name = "rx", type = "input", default = 0 },
+]
+
+[[blocks]]
+name = "gpio"
+instances = 5 # 2 RPi, 2 Ard, 1 Pmod
+
+[[blocks.ios]]
```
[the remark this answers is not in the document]
marnovandermaas: Ok, it was part of the GPIO block but I've now changed the syntax to make it more clear.

rtl/system/sonata_system.sv
```diff
@@ -36,76 +34,52 @@ module sonata_system #(
   // General purpose input and output
   input  logic [GpiWidth-1:0]      gp_i,
   output logic [GpoWidth-1:0]      gp_o,
+  // 0: Raspberry Pi HAT g0-15
+  // 1: Raspberry Pi HAT g16-27
+  // 2: Arduino Shield 0-15
+  // 3: Arduino Shield 16-17
+  // 4: PMOD
+  input  logic [HalfWordWidth-1:0] gp_headers_i [sonata_pkg::GPIO_NUM],
+  output logic [WordWidth-1:0]     gp_headers_o [sonata_pkg::GPIO_NUM],
```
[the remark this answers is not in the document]
marnovandermaas: This is now done with the new IP block.

rtl/fpga/top_sonata.sv
```diff
@@ -222,90 +224,120 @@ module top_sonata (
...
-  // Inputs from I2C buses.
-  wire scl0_i = scl0 & rph_g1;
-  wire sda0_i = sda0 & rph_g0;
+  logic [31:0] gpio_h2d_combined[GPIO_NUM];
 
-  wire scl1_i = scl1 & rph_g3_scl & mb6;
-  wire sda1_i = sda1 & rph_g2_sda & mb5;
+  for (genvar i = 0; i < GPIO_NUM; i++) begin : gen_gpio_combined
+    assign gpio_h2d_combined[i][31:16] = gpio_en_h2d[i];
+    assign gpio_h2d_combined[i][15: 0] = gpio_h2d[i];
```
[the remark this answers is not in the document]
marnovandermaas: Yes, I've removed this now based on the new GPIO unit with enables.

rtl/templates/pinmux.sv.tpl
```diff
@@ -0,0 +1,107 @@
...
+  assign ${pin_output}${idx_str}_sel = 'b10;
+
+  prim_onehot_mux #(
+    .Width(1),
+    .Inputs(${len(possible_blocks)+1})
+  ) ${pin_output}${idx_str}_mux (
+    .clk_i,
+    .rst_ni,
+    .in_i({
+      1'bz,
```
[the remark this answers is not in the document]
marnovandermaas: Agreed, this also allows the pinmux to be built using Verilator which is nice.

rtl/templates/pinmux.sv.tpl
```diff
@@ -0,0 +1,107 @@
...
+  input logic rst_ni,
+
+  // Blocks
+  % for label, width, name, instances in block_ios:
+  ${label} logic ${width}${name}[${instances}],
+  % endfor
+
+  // Pins
+  % for width, name in pin_ios:
+  inout logic ${width}${name},
```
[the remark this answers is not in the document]
marnovandermaas: I agree I've essentially merged the padring and the pinmux. I'm not sure if splitting it up will make a cleaner design, but I'm not against it. Would it be ok to merge the PR as is and then potentially split the padring and the pinmux in a future PR?

rtl/system/pinmux.sv
marnovandermaas: I'm not against having a way to generate registers, although Hugo mentioned an alternative SystemRDL as an alternative. Either way I think this is something we can resolve in a follow-up PR.

marnovandermaas: It's also mentioned as a task in the pinmux refinement issue: https://github.com/lowRISC/sonata-system/issues/169

rtl/templates/pinmux.sv.tpl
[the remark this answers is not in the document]
marnovandermaas: It's also mentioned in the pinmux refinement issue: https://github.com/lowRISC/sonata-system/issues/169

rtl/system/pinmux.sv
GregAC: Ultimately I think we should have consistency. Right now the OpenTitan reggen is probably the quickest way to consistency. I don't want to introduce another register interface generator so if we're going to use SystemRDL we'd need to switch over wholesale which isn't something we can support right now.

sw/cheri/checks/cheri_sanity.cc
```diff
@@ -21,14 +21,16 @@ using namespace CHERI;
...
 
-  asm volatile(
-      // Store capability to stack
-      "csc              cra, -4(csp)\n"
-      // Load capability from stack
-      "clc              cra, -4(csp)\n"
-      // Clear capability from stack
-      "csc              cnull, -4(csp)\n" ::
-          : "memory");
+  // TODO fix illegal instruction exception here.
```
HU90m, line 24: Remember to open an issue for this

rtl/system/pinmux.sv
```diff
@@ -0,0 +1,7962 @@
...
+      1'b0
+    }),
+    .sel_i(gpio_ios_2_31_sel),
+    .out_o(gpio_ios_o[2][31])
+  );
+
+  // Combining inputs for combinable inouts
+  assign i2c_scl_o[0] =
+    scl0 &&
+    rph_g1;
```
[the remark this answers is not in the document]
GregAC: Oh and this makes more sense as a bit-wise AND I think

rtl/templates/pinmux.sv.tpl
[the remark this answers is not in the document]
GregAC: It's about separation of concerns. The functionality of a pinmux is distinct from a pad instance/pad ring so it's better to de-couple them. It's particularly critical in this instance for two reasons:

1. Use of `inout` and `Z`. As I said in my earlier comment these need to be carefully handled in RTL. The pinmux is a large auto-generated module so there's lot to look at and think about so don't want to add ensuring correct usage of tristate as part of this. By keeping the pinmux strictly digital logic and isolating the tristate stuff into a per pad module there's a far smaller bit of RTL to review for correct tristate usage and lower maintenance burden going forward (we only have to think about tristate when we touch the pad instance). It would have reduced the effort required on my part to review this for instance.

2. Ease of switching pad implementation. Different targets (different FPGAs, ASIC etc) may want different pad instances. The generic `assign pad = pad_oe ? pad_o : 'z`  does do the job for FPGA most of the time but there may be reasons you want a technology specific pad instance and it would be essential were this ever to be used against an ASIC target.

Happy for this to get merged without this getting changed but I don't want it dropped as a 'nice-to-have' task that doesn't get done.

GregAC (approved):
Thanks for your work on this @marnovandermaas I think we're good to go once the issue with the I2C signal combine has been addressed.

doc/ip/pinmux.md
```diff
@@ -2,11 +2,21 @@
...
 
-We will use [the pin multiplexer from OpenTitan](https://opentitan.org/book/hw/ip/pinmux/index.html), but then parameterized for Sonata.
-More details on how this changes the register map will be provided in the future.
+To see the possible mappings, please refer to [the top configuration](https://github.com/lowRISC/sonata-system/blob/main/data/top_config.toml).
 
-In general, we expect each pin can be connected to a limited subset of blocks.
-The output of GPIO and PWM should only be able to control one pin, while blocks like UART, SPI and I2C should have multiple pins that they can connect to.
-The exact mapping of which pin can be driven by which hardware IP block will be provided in the future.
+Register addresses `0x000` to `0x7ff` are pin selectors, which are packed 4 per 32-bit word.
+Until we have an automated way to generate documentation, you'll have to look in [the generated pinmux RTL](https://github.com/lowRISC/sonata-system/blob/main/rtl/system/pinmux.sv)
```
HU90m, line 9: Once this is in and settled, we should document the default configuration. Mention of other top configurations should probably be left out of the reference manual and be somewhere else, somewhat hidden from normal users.

marnovandermaas: The table for this is automatically generated now and it mentions the default selected value is `'b10`

rtl/system/pinmux.sv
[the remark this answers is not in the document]
marnovandermaas: Good spot! I've added the selectors in here now and changed to bitwise operators.

sw/cheri/checks/cheri_sanity.cc
marnovandermaas: Here you go:
https://github.com/lowRISC/sonata-system/issues/206

rtl/system/pinmux.sv
```diff
@@ -0,0 +1,7962 @@
...
+// rtl/system/pinmux.sv is automatically generated using util/top_gen.py from rtl/templates/pinmux.sv.tpl
+// Please make any edits to the template file.
+
+module pinmux (
+  // Clock and reset.
+  input logic clk_i,
+  input logic rst_ni,
+
+  // List of block IOs.
+  input  logic uart_tx_i[5],
```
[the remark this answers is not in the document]
marnovandermaas: Captured in: https://github.com/lowRISC/sonata-system/issues/169

PR #156, opened 2024-08-19
Implement memset in bootloader and use it to reset revocation bitmaps

nbdd0121:
Also change `hardware_revoker_control_reg` to reset with the core instead of system.


sw/cheri/boot/boot_loader.cc
```diff
@@ -14,6 +14,11 @@
 #include <platform-uart.hh>
 #include <stdint.h>
 
+extern "C" {
+	// Use a different name to avoid resolve to CHERIoT-RTOS memset symbol.
+	void bl_memset(void *, int, size_t);
+}
```
HU90m, line 20: :crab:

HU90m (approved):

PR #157, opened 2024-08-20
HyperRAM fix and improvements

nbdd0121:
Two changes made:
* HyperRAM controller tag RAM's cmd and wdata FIFO get merged into one. There's only 1 bit for the data so there isn't much need to separate cmd and data FIFO.
* Initialise `ufifo_rd_ena` in the beginning of `always_comb` block to zero. Otherwise Vivado will infer it as a latch.


GregAC:
Good job spotting this @nbdd0121 and thanks for the FIFO merge, it certainly simplifies the controller design.

I should probably get the lint job running on the hyperram controller! (It can't lint it all due to the Xilinx IP used but could at least lint this file with other stuff blackboxed out).

GregAC (approved):

PR #158, opened 2024-08-20
Arbitrary entrypoint support in bootloader

nbdd0121:
Currently we jump to a hardcoded address, but we can actually take the value from the ELF header instead.

This PR also include commits that:
* add an exception handler, for better debugging when things go wrong
* allow bootloader to load into HyperRAM address range in addition to SRAM


sw/cheri/boot/boot_loader.cc
```diff
@@ -128,13 +153,42 @@ extern "C" void rom_loader_entry(void *rwRoot)
...
 	spi_flash.reset();
-	read_elf(spi_flash, uart, sram);
+	uint32_t entrypoint = read_elf(spi_flash, uart, sram, hyperram);
 
 	write_str(uart, prefix);
 	write_str(uart, "Booting into program, hopefully.\r\n");
+	return entrypoint;
+}
+
+#define READ_CSR(name) ({ \
```
HU90m, line 172: Please could you add a comment pointing towards https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html

HU90m (approved):

PR #159, opened 2024-08-21
Fix various minor Synth warnings

elliotb-lowrisc:
Good to keep on top of warnings.


elliotb-lowrisc:
I'm not super sure about the gpio.sv change, given how verbose it is. I can remove it from this PR if requested

nbdd0121:
Is gpio.sv change to ignore the 0 width replication warning? I think we should just allow that warning. It's allowed in SV spec and I don't think thee verbosity worths it.

elliotb-lowrisc:
> Is gpio.sv change to ignore the 0 width replication warning? I think we should just allow that warning. It's allowed in SV spec and I don't think thee verbosity worths it.

It is, shall I remove that commit then?
I wish there was a way to sign-off warnings or some other way of hiding them from a report of warnings

nbdd0121:
I would err on the side of removing the commit.

Alternatively if we want to appease vivado, we can change the assignment line to `DataWidth'(gp_o)` without having to use generate if, but then we lose the ability to detect truncation.

vendor/lowrisc_ibex/vendor/lowrisc_ip/ip/prim/rtl/prim_util_memload.svh
HU90m, line 1: The vendored IP would need to fixed in a patch applied by the vendor script or fixed upstream. *This piece is vendored twice, so is originally from opentitan.*

elliotb-lowrisc: Does that mean I shouldn't change it?

HU90m: Probably easier not to change this particular one. Thanks for working on this though, really appreciated.

nbdd0121: BTW isn't `$readmemh` needed for synthesis?

elliotb-lowrisc: Presumably not in Vivado, given that it was ignoring the whole initial block (based on the warning message - see commit).

nbdd0121: I think it could be a misleading diagnostics. This seems to be the only effective `$readmemh` for synthesis, and gating this would mean that the BRAMs are not initialised. `initial` block with `$readmemh` is supported by Vivado, it might just be the first if condition being ignored. Have you tested this PR with `--SRAMInitFile`?

elliotb-lowrisc: Ah, so you think it's just ignoring line 62, and not the later ones. 

No I haven't tested it. How do I?

nbdd0121: https://github.com/lowRISC/sonata-system/blob/main/sw/cheri/README.md#embedding-software-into-a-bitstream

elliotb-lowrisc:
I've removed the replication bit change and narrowed the scope of the memory init change to just the memory path printing bit that Vivado was complaining about. Tested with:
```
fusesoc --cores-root=. run --target=synth --setup --build lowrisc:sonata:system --SRAMInitFile=/home/eb/sonata-system/sw/cheri/build/tests/cheri_sanity.vmem
openFPGALoader -c ft4232 build/lowrisc_sonata_system_0/synth-vivado/lowrisc_sonata_system_0.bit
```

GregAC (approved):
Thanks for the cleanup @elliotb-lowrisc LGTM.

Do you have that bit replication fix you still did anywhere? I'd like to take a look at it even if we don't want to merge it in as you say good to clean up warnings and I didn't get a chance to see how verbose/awkward it was myself.

nbdd0121:
https://github.com/lowRISC/sonata-system/commit/d30ea06c662144c67868d880beeb2682cbfa8311

nbdd0121:
I think Hugo's point about the vendored IP is yet to be addressed; the changes would be overwritten when someone rerun the vendor script.

elliotb-lowrisc:
> I think Hugo's point about the vendored IP is yet to be addressed; the changes would be overwritten when someone rerun the vendor script.

Ah, I lost track of that. I think this PR has taught me not to group commits into the same PR like this.

I shall have a go at changing this in OpenTitan then.

vendor/lowrisc_ibex/vendor/lowrisc_ip/ip/prim/rtl/prim_util_memload.svh
elliotb-lowrisc: I've created an OpenTitan PR: https://github.com/lowRISC/opentitan/pull/24400

Seems to be blocked for some reason at present

PR #161, opened 2024-08-21
Add post-synth hook & split XDC

elliotb-lowrisc:
Constraints such as false_paths cannot be applied
until after elaboration/synthesis, such as internal false paths. Edalize/fusesoc do not seem to provide a mechanism for this, so use the `tclSource` and `user` filetypes they do provide and the pre/post stage hook mechanism provided by Vivado.

Split the existing XDC into two files. The only real change being that some of the new HyperRAM `set_false_path` constraints are moved to a new "paths_sonata.xdc" file that is read post-synthesis. Issues with these moved constraints likely remain, but at least they have a better chance of working.


nbdd0121:
You can set a file to be used in synthesis or not using
```tcl
set_property used_in_synthesis false [get_files data/paths_sonata.xdc]
```

This would avoid user file_type and copies.

elliotb-lowrisc:
> You can set a file to be used in synthesis or not using
> 
> ```tcl
> set_property used_in_synthesis false [get_files data/paths_sonata.xdc]
> ```
> 
> This would avoid user file_type and copies.

Oh, I thought we were using Non-Project mode. Thanks, I'll try `used_in_synthesis`.

elliotb-lowrisc:
It seems the `used_in_synthesis` option works, but the resulting `read_xdc` command lacks the `-unmanaged` argument that would be required for the commented-out `foreach` commands to work. @GregAC would this be a limitation you would be happy with?

nbdd0121:
According to https://docs.amd.com/r/en-US/ug903-vivado-using-constraints/About-XDC-Constraints, this might work to make it unmanaged:
```tcl
set_property file_type Tcl [get_files data/paths_sonata.xdc]
```

GregAC:
> It seems the `used_in_synthesis` option works, but the resulting `read_xdc` command lacks the `-unmanaged` argument that would be required for the commented-out `foreach` commands to work. @GregAC would this be a limitation you would be happy with?

I would like those foreach  commands to be usable eventually, even if it's not a high priority now. Though given we're already refactoring this area best set things up so we can use them even if they're left commented out for now. I don't think the user filetype and the copyto are that big of an issue, maybe add a comment into the core file to explain what's happening with them. Though if @nbdd0121's suggestion works we can go with that too.

elliotb-lowrisc:
> According to https://docs.amd.com/r/en-US/ug903-vivado-using-constraints/About-XDC-Constraints, this might work to make it unmanaged:
> 
> ```tcl
> set_property file_type Tcl [get_files data/paths_sonata.xdc]
> ```

Thanks, that worked. It seems I need to adjust my way of thinking from they phys custom-flow way I'm familiar with to this managed flow.

I'll update this PR shortly

GregAC:
> > According to https://docs.amd.com/r/en-US/ug903-vivado-using-constraints/About-XDC-Constraints, this might work to make it unmanaged:
> > ```tcl
> > set_property file_type Tcl [get_files data/paths_sonata.xdc]
> > ```
> 
> Thanks, that worked. It seems I need to adjust my way of thinking from they phys custom-flow way I'm familiar with to this managed flow.
> 
> I'll update this PR shortly

This is FPGA land, trying to reduce warnings and actually have decent timing constraints is just a bizarre thing no-one does ;)

Good work on this, very glad to see these kinds of improvements going in.

elliotb-lowrisc:
I mean, if you don't care about reliability, then ignore warnings and timing constraints all you want :)

elliotb-lowrisc:
Note that I have split the XDC file into three files now, as per [UG949](https://docs.amd.com/r/en-US/ug949-vivado-design-methodology/Simple-Design): "1 file for physical + 1 file for timing (synthesis) + 1 file for timing (implementation)"

nbdd0121 (approved):

PR #162, opened 2024-08-22
Add XADC

elliotb-lowrisc:
Add a generated wrapper for the XADC built-in to the FPGA and an adapter to connect it to the TLUL crossbar
via the XADC's Dynamic Reconfiguration Port (DRP).

The DRP registers are all 16-bits, and have been mapped at 32-bit intervals in the TLUL address space.
All writes must be 16-bit writes or they will be silently dropped. Any writes outside of the range of writable registers will be silently dropped.
Any reads above the range of DRP registers will cause an error.

Driver being worked on by @AlexJones0

### DIFF
--- a/rtl/ip/xadc/rtl/xadc.sv
+++ b/rtl/ip/xadc/rtl/xadc.sv
@@ -1,0 +1,164 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+module xadc (
+  input clk_i,
+  input rst_ni,
+
+  input  tlul_pkg::tl_h2d_t tl_i,
+  output tlul_pkg::tl_d2h_t tl_o,
+
+  // Analog(ue) input width not parametrised due to PCB-specific
+  // mapping to XADC pins (see u_xadc below).
+  input wire [5:0] analog_p_i,
+  input wire [5:0] analog_n_i
+);
+
+  // TLUL parameters
+  // Specify the 4-byte address width required to cover the whole of the
+  // address space assigned to the XADC in the TLUL crossbar.
+  // That way, `xadc_adapter` is not susceptible to aliasing.
+  localparam int TAW = 10;
+  localparam int TDW = 32;
+
+  // DRP 'parameters' (fixed by XADC Hard-IP) to aid readability
+  localparam int XAW = 7;
+  localparam int XDW = 16;
+
+  logic           drp_dclk;
+  logic           drp_den;
+  logic           drp_dwe;
+  logic [XAW-1:0] drp_daddr;
+  logic [XDW-1:0] drp_di;
+  logic           drp_drdy;
+  logic [XDW-1:0] drp_do;
+
+  // TLUL to DRP adaptor specifically for XADC
+  xadc_adapter #(
+    .TAw(TAW),
+    .TDw(TDW),
+    .XAw(XAW),
+    .XDw(XDW)
+  ) u_xadc_adapter (
+    .clk_i       (clk_i),
+    .rst_ni      (rst_ni),
+
+    .tl_i        (tl_i),
+    .tl_o        (tl_o),
+
+    .drp_dclk_o  (drp_dclk),
+    .drp_den_o   (drp_den),
+    .drp_dwe_o   (drp_dwe),
+    .drp_daddr_o (drp_daddr),
+    .drp_di_o    (drp_di),
+    .drp_drdy_i  (drp_drdy),
+    .drp_do_i    (drp_do)
+  );
+
+`ifndef SYNTHESIS
+  // Fake XADC - responds with the number of requests submitted
+
+  // Respond the cycle after a request
+  // TODO: randomise response delay for better bug-finding
+  logic respond_d;
+  logic respond_q;
+  assign respond_d = drp_den;
+  always_ff @(posedge drp_dclk or negedge rst_ni) begin
+    if (!rst_ni) begin
+      respond_q <= '0;
+    end else begin
+      respond_q <= respond_d;
+    end
+  end
+
+  // Increment counter when get a request, and use count as response data
+  logic [XDW-1:0] req_count_d;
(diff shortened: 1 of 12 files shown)


GregAC (commented):
Did you consider using reggen to generate the top-level tile link interface? Documentation is here: https://opentitan.org/book/util/reggen/index.html and the actual tool is in the opentitan repository.

Though given the DRP takes multiple cycles to fulfill a request we unfortunately can't just connect things straight-through which is maybe why you've got this design?

We actually have a similar issue in OpenTitan which reggen provides functionality for where the register access needs to cross through to a slower clock domain (e.g. used in aon_timer).

Here a write returns an immediate response and then makes it way over to the slower clock domain. Reads wait until actual data is available.

What you get from reggen for this CDC case won't map directly to what you need here as it handles all of the waiting etc internally. However the `tlul_adapter_reg` module does have a 'busy_i' input to support this behaviour. Could be by generating something with reggen that gives full hardware control to the read/write and then a little manual mod to the autogenerated files to expose the `busy_i` input of all the registers will work for this use case.

That said it's potentially quite a bit of work to get that working but I'm also not a fan of using custom logic here rather than reusing our existing known good primitives. Plus if we use reggen we can also auto-generate register documentation and C headers (which we don't do right now but it's useful to have the capability).

Another option would be rather than exposing the actual registers as tilelink addressable things just expose a register interface that allows you to access the DRP address space e.g. with 4 registers 'address', 'write_data', 'read_data', 'control', 'status'. To read you write the DRP address you need in 'address' and then set a 'read' bit in 'control' then poll 'status' until it indicates done and read from 'read_data'. For write set 'address' and 'write_data' and then set the 'write' bit in 'control' and you can poll a 'busy' bit in 'status' to wait for completion if you want. Any register access would be ignored when 'busy' is set in 'status'.

A more awkward software interface but easier RTL to write here and simpler means less chance of bugs.

rtl/ip/xadc/rtl/xadc.sv
```diff
@@ -0,0 +1,149 @@
...
+    .drp_daddr_o (drp_daddr),
+    .drp_di_o    (drp_di),
+    .drp_drdy_i  (drp_drdy),
+    .drp_do_i    (drp_do)
+  );
+
+`ifndef SYNTHESIS
+  // Fake XADC - responds with the number of requests submitted
+
+  logic pavlova_drdy_d;
```
GregAC, line 62: pavlova?

rtl/ip/xadc/rtl/xadc.sv
```diff
@@ -0,0 +1,149 @@
...
+    .drp_dclk_o  (drp_dclk),
+    .drp_den_o   (drp_den),
+    .drp_dwe_o   (drp_dwe),
+    .drp_daddr_o (drp_daddr),
+    .drp_di_o    (drp_di),
+    .drp_drdy_i  (drp_drdy),
+    .drp_do_i    (drp_do)
+  );
+
+`ifndef SYNTHESIS
```
GregAC, line 59: Not for this PR but we should do a little model of the DRP interface for synthesis that returns some value after a random time, this will allow us to use the adapter logic in simulation which will be useful in finding/fixing any bugs here.

rtl/ip/xadc/rtl/xadc_adapter.sv
```diff
@@ -0,0 +1,153 @@
...
+  logic above_addr_range;
+  // All accesses should be to an address offset inside the
+  // range of the 128 (0-127) XADC DRP registers.
+  assign above_addr_range = |mem_addr[TAw-1:XAw];
+
+  // Write-specific request issue detection
+  logic wmask_bad, wr_to_rd_only_addr;
+  // Writes should be to the lower 16-bits (no more & no less) in this simple
+  // implementation, as the underlying XDC registers are all 16-bit registers.
+  assign wmask_bad = mem_we && (~&mem_wmask[XDw-1:0] || |mem_wmask[TDw-1:XDw]);
```
GregAC, line 115: I'd allow full 32-bit writes as well as lower 16-bit writes. In general all device memory accesses are done on aligned 32 bit accesses, so this would be surprising special case behaviour, especially as it would just silently ignore 32-bit writes.

rtl/ip/xadc/rtl/xadc.sv
elliotb-lowrisc: Just some random word. I thought "foo" would be a bit boring. I suppose something like "fake" or "mock" could also work.

rtl/ip/xadc/rtl/xadc.sv
elliotb-lowrisc: :+1:

rtl/ip/xadc/rtl/xadc_adapter.sv
elliotb-lowrisc: Alright, I'm happy to change this if it would be a better fit for the system. It sounds like it may have been causing problems on the software side.

elliotb-lowrisc:
Dropped write mask upper bits clear requirement, tweak "fake XADC" signal names, unused bits assignment, changed XADC template for slower default sample clock and excluding unconnected analogue inputs.

Apologies for all the pushes. Should be fewer in future now that I've found how to run more linting locally.

AlexJones0 (commented):
I can't check the RTL, but just wanted to note that as of the most recent force-push, this seems to now be working on the software side. Building the bitstream for this PR (plus, only tested with CHERI disabled for now) and testing on FPGA, I can confirm that I can read/write the DRP registers mapped through MMIO. Inputs connected via the Arduino Analogue pins on the Sonata board can be seen varying through the mapped sampled status registers, and writing to the XADC config registers does appear to impact the operation of the XADC.

rtl/ip/xadc/rtl/xadc.sv
elliotb-lowrisc: Now renamed to something more descriptive

marnovandermaas (approved):
Just did a review of this code and it looks good to me. In combination with the fact that Alex says that it works in software, I think it is reasonable to approve this PR.

PR #165, opened 2024-08-26
[util] Fix sonata-openocd-cfg.tcl for use with OpenOCD 0.12.0

GregAC:
The set_prefer_sba command has been deprecated, set_mem_access should be used instead.


marnovandermaas (approved):
Seems reasonable to me.

PR #166, opened 2024-08-27
Added license lint

HU90m:


.reuse/dep5
```diff
@@ -0,0 +1,47 @@
...
+
+Files: vendor/newae/crypto/tiny-AES128-C/*
+Copyright: tiny-AES-c contributors
+License: Unlicense
+
+Files: vendor/newae/simpleserial*
+       vendor/newae/hal/hal.h
+       vendor/newae/basic-passwdcheck/basic-passwdcheck.c
+Copyright: 2012-2017 NewAE Technology Inc
+License: GPL-3.0-or-later
```
nbdd0121, line 39: Looks like we have some license compatibility issue. We should either remove the dependency, or ask NewAE to relicense them to a permissive license.

HU90m: Please can we still get this PR in to avoid more licensing issues in the future.

*@GregAC for visibility into the licensing problem*

marnovandermaas:
Can you point me to the code that actually does the lint? I think it would probably be good to split this into two commits (one with the license changes and one with the CI changes).

marnovandermaas (approved):
Yes this should go in thank you. Approving so that you can rebase and then merge.

PR #168, opened 2024-08-29
USB DPI model integration

alees24:
Integrate the USBDPI host model into the Sonata System for use in simulations.

Modify the common/usbdev driver and hello_usb code to support the vendor specific test description that the USB DPI model requires. This permits demos/hello_usb to continue executing in simulation, although presently there is no attempt to log the output from this FPGA-based test program.

The operation of the USB DPI model may be checked at present by disabling CHERI in dv/verilator/top_verilator.sv and then executing `sw/legacy/build/demo/hello_usb/hello_usb`. That is a FPGA-based test program that does not complete and expects the USB host to collect serial data (a la hello_world); the DPI model does not do that presently but it will run the USB device and software through the normal configuration sequence and the simulation runs indefinitely, as per the FPGA test.
A CHERI-enabled test application is under construction; to be included later.


marnovandermaas (approved):
Looks good to me from a code inspection. Assuming you have tested the code feel free to merge.

PR #172, opened 2024-08-30
Icache

GregAC:


dv/verilator/sonata_system.cc
```diff
@@ -40,6 +42,7 @@ int SonataSystem::Setup(int argc, char **argv, bool &exit_app) {
                  VerilatorSimCtrlFlags::ResetPolarityNegative);
 
   _memutil.RegisterMemoryArea("ram", 0x100000, &_ram);
+  _memutil.RegisterMemoryArea("hyperam", 0x40000000, &_hyperram);
```
nbdd0121, line 45:
```suggestion
  _memutil.RegisterMemoryArea("hyperram", 0x40000000, &_hyperram);
```

sw/cheri/sim_boot_stub/boot.S
```diff
@@ -5,7 +5,7 @@
 	.p2align 2
     .type start,@function
 start:
-	auipcc           cra, 0
-	li               t0, 0x00101000
-	csetaddr         cra, cra, t0
-	cjr              cra
+  auipcc           ct1, 0
```
nbdd0121, line 10: Currently we have tab indentation for all .S files. We should decide on a style and reformat all .S files.

vendor/lowrisc_ibex.lock.hjson
```diff
@@ -1,4 +1,4 @@
-// Copyright lowRISC contributors.
+// Copyright lowRISC contributors (OpenTitan project).
```
nbdd0121, line 1: Is this change intended?

sw/cheri/tests/hyperram_tests.hh
```diff
@@ -217,6 +217,8 @@ void write_prog(Capability<volatile uint32_t> hyperram_area, uint32_t addr) {
   hyperram_area[addr + 2] = 0x00552023;
   hyperram_area[addr + 3] = 0x00000517;
   hyperram_area[addr + 4] = 0x8082;
+
+  asm volatile ("fence.i");
```
nbdd0121, line 221: Technically not needed because `hyperram_area` is volatile, but I think it's still worth adding to capture that memory op reordering must not happen for this instruction:

```suggestion
  asm volatile ("fence.i" ::: "memory");
```

sw/cheri/sim_boot_stub/boot.S
GregAC: We should follow the lowRISC style guide as we do for all other lowRISC repositories, so space indentation. Yes we do need to reformat this repository.

I think we may make an exception for sonata-software as it's more important for it be compatible with CHERIoT RTOS upstream which uses tabs.

vendor/lowrisc_ibex.lock.hjson
GregAC: I think it comes from the OpenTitan vendor.py script, so not quite correct but I think important we leave the files are they were generated from vendor tooling, we can maybe add some flag to that tool so you suppress this addition.

GregAC:
Note to test hyperram execution with icache you'll need to use this branch: https://github.com/lowRISC/cheriot-rtos/tree/hyperram of our CHERIoT RTOS fork inside sonata-software.

Before we do the v0.4 release we'll need to update the sonata-software submodule to point to that branch and make a v0.4 branch in sonata-software for use with the release.

marnovandermaas (approved):
Dropping the Verilator changes but otherwise I've tested this on FPGA and it works.

PR #174, opened 2024-08-30
Add fence after writing program with icache

marnovandermaas:


GregAC:
Ah missed this was still in checks/ as well. We should really unify the code here so both the tests/ and checks/ versions are pulling from the same source (the checks/ version will just loop forever and the tests/ version will just run once).

GregAC (approved):

PR #181, opened 2024-09-05
Simple bare metal CHERI check application for USB device.

alees24:
This does not require any special software on a physical USB host controller; `usbdev_check` will simply set up and connect the USB device to the USB host controller and await successful completion of the the normal device inspection and configuration sequence by the host.

If the `do_disconnect` switch in usbdev_check.cc is set to false then serial comunications may also be tested by opening a terminal emulator on /dev/ttyUSB<n> where the number 'n' may be ascertained by monitoring the output from dmesg when the USB device is being configured.

Notes on the code:

 - There appear to be a couple of coding styles in use. I'm aware that platform-usbdev.hh presently uses spaces rather than tabs and will not be accepted upstream. It's anticipated to change before upstreaming anyway.
 - usbdev_utils.hh forms an intermediate layer between the `usbdev_check.cc` simple application code. This layer handles the configuration sequence over the 'Default Control Pipe'. It can form the basis of a more complete test application.
 - Indeed if the switch 'do_disconnect' is changed to false then the application will keep running and provide a basic ttyUSB<> serial port via the `simpleserial` driver.
 - I've tried to mimic formatting, naming conventions etc to aid the development of this code into a CHERIoT-RTOS based application consisting of multiple compartments as previously oulined.


sw/cheri/checks/usbdev_check.cc
```diff
@@ -0,0 +1,160 @@
...
+}
+
+static void uart_logn(Capability<volatile OpenTitanUart> &uart, const char* str, size_t len)
+{
+  while (len-- > 0u)
+  {
+    uart->blocking_write(*str);
+    ++str;
+  }
+}
```
marnovandermaas, line 46: Can't you use `write_str` instead?
https://github.com/lowRISC/sonata-system/blob/275eaab74ebec4fcf045b071eb834e586d1d28f5/sw/cheri/common/uart-utils.hh#L13

marnovandermaas (approved):
Just one comment on using the uart-utils or if you don't have what you need there then add it there instead of in individual tests. Other than that happy for this PR to be merged. Once you have made the UART changes feel free to merge.

alees24:
I was mimicking uart_check.cc which contained its own implementation; now changed. It may be worth mentioning that including a header file which contains functions defined as 'static' can lead to warnings from the compiler if a translation unit happens not to use all of the functions in the header file. The builds are okay for now.

PR #182, opened 2024-09-09
Add a lightweight implementation of ostream

engdoreis:
Add a lightweight implementation of ostream that can be easily extended for other types and formats.


marnovandermaas:
> I'd prefer us to not use ostream. If possible we should use `std::print` equivalent, otherwise I would prefer printf to ostream.

Why do you not like `ostream`?

nbdd0121:
Because it's bad enough that C++23 introduces a completely new Rust-like print function? https://en.cppreference.com/w/cpp/io/print

engdoreis:
Closing in favour of https://github.com/lowRISC/sonata-system/pull/229

PR #183, opened 2024-09-10
Add support for software slots

AlexJones0:
This PR implements support for software slots in the RTL and bootloader. This incorporates the existing UF2 build flow - UF2 files must be built with the correct target address offsets, e.g. `0x0000_2000` for slot 1, `0x1000_2000` for slot 2, and `0x2000_2000` for slot 3. Software slot is specified by the top nibble in the UF2's target address.

The bootloader then reads the software select GPIO to determine which program to load from flash.

SPI flash reads have been changed to use 32-bit addresses instead of 24-bit addresses, as I had to spend a couple of hours debugging incorrect reads of software slot 3 due to the silent dropping of the upper byte of addresses.

This has been tested by building the bitstream with the new bootloader, and then creating 3 different pieces of demo software with the offsets listed above. Loading each of them, the software selection switch then works as expected, with a simple Reset (`SW5`) causing the new firmware to load after changing the selected slot. No power cycling is needed.


sw/cheri/boot/boot_loader.cc
```diff
@@ -62,6 +92,7 @@ static void write_hex_with_prefix(UartRef uart, const char *msg, uint32_t value)
 }
 
 uint32_t read_elf(SpiFlash                  &flash,
+				  SoftwareSlotAddress        addr,
```
nbdd0121, line 95: I wonder rather see this as a uint32_t directly to avoid the casts inside this function. Cast can happen before calling `read_elf`.

sw/cheri/boot/boot_loader.cc
```diff
@@ -40,6 +58,18 @@ typedef CHERI::Capability<volatile OpenTitanUart> &UartRef;
 	}
 }
 
+uint8_t read_selected_software(GpioRef gpio)
```
nbdd0121, line 61:
```suggestion
size_t read_selected_software_slot(GpioRef gpio)
```

sw/cheri/boot/boot_loader.cc
```diff
@@ -26,9 +26,27 @@ extern "C" {
...
+	Slot1 = 0 * 10 * 1024 * 1024,
+	Slot2 = 1 * 10 * 1024 * 1024,
+	Slot3 = 2 * 10 * 1024 * 1024,
+};
+
+const SoftwareSlotAddress SoftwareSlots[] = {
+	SoftwareSlotAddress::Slot1,
+	SoftwareSlotAddress::Slot2,
+	SoftwareSlotAddress::Slot3
+};
```
nbdd0121, line 42: I think we don't need this additional enum class, as this causes churn later. Normal enums or just addresses directly in `SoftwareSlots` is good enough.

nbdd0121 (approved):

PR #185, opened 2024-09-13
Format the code base using clangformat

engdoreis:
Use the same configuration in ibex-demo-system.


marnovandermaas (commented):
Thanks for putting this together, can you change the CI to ignore C files in vendor directory?

I see a lot of warnings under the clang format tab:
```
Run find -name "*.[cc|c|h]" -exec clang-format -n {} \;
./vendor/display_drivers/core/m3x6_16pt.c:966:6: warning: code should be clang-formatted [-Wclang-format-violations]
    { 5, 0 },  // ' '
```

HU90m (changes requested):
Thanks for doing this, looks broadly good. Would be nice to have the check in `lint-all`. Also would be great to follow the name convention `lint-cpp` so that additional c++ lints (namely `clang-tidy`) can be added to the same lint in the future. See sonata-software for prior art

Also, I started to move away from. Providing fix applications. We can discuss and I can dig up the prs where I removed them for lints in the sonata-* repos

engdoreis:
@HU90m I think that I have addressed all the comments.

flake.nix
```diff
@@ -74,8 +74,20 @@
         ${getExe pkgs.reuse} --suppress-deprecation lint
         ${getExe pkgs.lychee} --offline --no-progress .
         ${getExe lint-python}
+        ${getExe format-cpp}
       '';
 
+      format-cpp = pkgs.writeShellApplication {
```
HU90m, line 80: It would be nice if this included followed the naming conventions of sonata-software [`lint-cpp`](https://github.com/lowRISC/sonata-software/blob/72e04e7ef29afe0ee3d1833eb0ac4be0bf765981/flake.nix#L143). I'd like to add `clang-tidy` in the future.

flake.nix
```diff
@@ -74,8 +74,20 @@
...
+        name = "format-cpp";
+        runtimeInputs = [pkgs.clang-tools_18];
```
HU90m, line 82: We should really be using the clang-format from the CHERIoT toolchain.

```suggestion
        runtimeInputs = with lrPkgs; [llvm_cheriot];
```

flake.nix
```diff
@@ -74,8 +74,20 @@
...
+        text = ''
+          set +u
+          ARG="$1"
+          [ -z "$1" ] && ARG="-n"
+          find sw -path sw/cheri/build -prune -o \( -name "*.c" -o -name "*.cc" -o -name "*.h" -o -name "*.hh" \) -exec clang-format "$ARG" {} \;
```
HU90m, line 87: The exclude path is a little hidden here. Just a note, I won't ask you to change this.

flake.nix
```diff
@@ -74,8 +74,20 @@
...
+        ${getExe format-cpp}
       '';
 
+      format-cpp = pkgs.writeShellApplication {
+        name = "format-cpp";
+        runtimeInputs = [pkgs.clang-tools_18];
+        text = ''
+          set +u
+          ARG="$1"
+          [ -z "$1" ] && ARG="-n"
```
HU90m, line 86: I'm not keen on this approach of passing through arguments directly. If people want to use clang-format argument, they can call the tool directly.

If we want to add an easy fix option we could do something similar to what we used to have with sonata-software: https://github.com/lowRISC/sonata-software/blob/206f88d8c81096c46c771ba80bb94c96bc3efa82/flake.nix#L71

HU90m (changes requested):

flake.nix
engdoreis: The clang-format from CHERIot toolchain seems to be using an old version that does not allow configuring the `AllowShortLoopsOnASingleLine`.
See details: https://github.com/llvm/llvm-project/issues/61708

flake.nix
engdoreis: Done

HU90m (approved):

PR #186, opened 2024-09-13
Implement SPI DPI and models of ST7735 LCD and W25Q256 flash

alees24:
This PR introduces a DPI implementation of SPI bus/devices.

A SystemVerilog module is parameterised for multiple chip selects, devices and out-of-band signals in each direction.
Sampling occurs on SCK rising edge, launching of data on SCK falling edge; if support for other clocking schemes is required it may be added later.
A software model of basic ST7735 LCD functionality produces a PPM file showing the image displayed on the LCD.
A simple model of the W25Q256 flash device that we use on the Sonata FPGA is included; it is sufficient for the `spi_test` code to produce the same output as on the physical system, but may require extension later.

It should be possible to add models of other SPI devices easily enough. Note that presently the code assumes a single SPI device per bus because this is how the current Sonata system is constructed, but it may require additional work later to support modelling of expansion cards, for example.

Testing: has been tested with `sw/legacy/demo/lcd_st7735/` (timer_delay modified to eliminate the delay) and `sw/legacy/test/spi_test`. If trying these, remember to set 'EnableCHERI' to 1'b0 in `dv/verilator/top_verilator.sv`


HU90m (approved):
This is great thanks Adrian! Drawing to the LCD is a little slower in sim but very fun.

Only requested change is that C++ headers are changed to `.hh`.

PR #187, opened 2024-09-16
Add I/O timing cons, fix speed-grade, faster JTAG

elliotb-lowrisc:
- Add I/O timing constraints for all ports.
- Correct FPGA speed-grade to the faster "-2" grade.
- Increase openocd JTAG speed using on-board programmer (used when loading programs into SRAM from computer).
- Add missing HyperRAM physical IOB constraint.
- Tidy pins_sonata.xdc and update deprecated properties.

The biggest change (by volume) is by far the I/O timing constraints. These should improve timing on some synchronous interfaces, such as the User JTAG, and improve reliability.
Further changes/refinements are likely.
Note that internal timing constraints have yet to be tackled.

Also note that build times may increase slightly due the there being more for the tool to consider and optimise timing-wise.

### DIFF
--- a/data/synth_timing.xdc
+++ b/data/synth_timing.xdc
@@ -4,29 +4,687 @@
 
 # This file is for timing constraints to be applied *before* synthesis.
 # i.e. timing constraints on top-level ports.
+#
+# See UG949 and UG903 for information on setting various timing constraints.
 
-## Clocks
-create_clock -period 40.000 -name mainClk -waveform {0.000 20.000} [get_ports mainClk]
-create_clock -period 100.000 -name tck_i -waveform {0.000 50.000} [get_ports tck_i]
+#### Recommended timing constraints sequence from UG949 ####
+## Timing Assertions Section
+# Primary clocks
+# Virtual clocks
+# Generated clocks
+# Delay for external MMCM/PLL feedback loop
+# Clock Uncertainty and Jitter
+# Input and output delay constraints
+# Clock Groups and Clock False Paths
+## Timing Exceptions Section
+# False Paths
+# Max Delay / Min Delay
+# Multicycle Paths
+# Case Analysis
+# Disable Timing
 
-## Clock Domain Crossings
-set clks_sys_unbuf  [get_clocks -of_objects [get_pin u_clkgen/pll/CLKOUT0]]
-set clks_usb_unbuf  [get_clocks -of_objects [get_pin u_clkgen/pll/CLKOUT1]]
 
-## Set asynchronous clock groups
-set_clock_groups -group ${clks_sys_unbuf} -group ${clks_usb_unbuf} -group mainClk -asynchronous
+#### Timing Assertions Section ####
+
+### Primary clocks ###
+create_clock -name mainClk -period 40.0   [get_ports mainClk] ;# 40 ns = 25 MHz
+# create_clock -name tck     -period 33.333 [get_ports tck_i]   ;# 33 ns = 30 MHz (FT4232HQ max, too fast for us?)
+create_clock -name tck     -period 66.666 [get_ports tck_i]   ;# 66 ns = 15 MHz (next step down from FT4232HQ max)
+
+### Virtual clocks ###
+create_clock -name vclk_extusb -period 83.333 ;# 83 ns = 12 MHz (full-speed USB)
+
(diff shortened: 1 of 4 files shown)


elliotb-lowrisc:
This will probably need reworking to account for pinmux once that is merged

elliotb-lowrisc:
Forgot multicycle constraint for Ethernet MAC SPI

elliotb-lowrisc:
Remove unused clock, add SPI clocks for R-Pi, loosen SPI input constraints a touch to ease implementation

data/synth_timing.xdc
```diff
@@ -4,29 +4,860 @@
...
 set_false_path -to [get_ports hyperram_cs]
 set_false_path -to [get_ports hyperram_nrst]
+
+### Max Delay / Min Delay ###
+## Ethernet MAC
+# Asynchronous(?) interrupt.
+# Use max_delay rather than false_path to keep the delay within reason.
+set_max_delay 30 -datapath_only -from [get_ports ethmac_intr]
+
+### Multicycle Paths ###
```
GregAC, line 687: Were any of these multi-cycle paths necessary to close timing? Perhaps we can get some better implementation results by giving the optimizer a easier job but I am very wary of including multi-cycle paths. In particular perhaps something is indeed multi-cycle right now but after some RTL change it isn't and these don't get appropriately adjusted.

So I'd probably drop them all unless there's specific cases you've found they're required. Will also be more complicated once pinmux is in there.

GregAC (commented):
Great work @elliotb-lowrisc! We'll be leading the way in open FPGA design quality with excellent timing constraints.

The one area I'm unsure about if the use of the multi-cycle paths, see my comment for more details.

data/synth_timing.xdc
```diff
@@ -4,29 +4,860 @@
...
+# create_clock -name tck     -period 33.333 [get_ports tck_i]   ;# 33 ns = 30 MHz (FT4232HQ max, too fast for us?)
+create_clock -name tck     -period 66.666 [get_ports tck_i]   ;# 66 ns = 15 MHz (next step down from FT4232HQ max)
+
+### Virtual clocks ###
+create_clock -name vclk_extusb -period 83.333 ;# 83 ns = 12 MHz (full-speed USB)
+
+### Generated clocks ###
+# PLL clocks - name only; period will be derived from RTL parameters.
+# All are generated from mainClk.
+create_generated_clock -name clk_sys   [get_pin u_clkgen/pll/CLKOUT0]
```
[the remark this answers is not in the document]
elliotb-lowrisc: However, we may be able to find the pin from the net name. I'll take a look.

data/synth_timing.xdc
elliotb-lowrisc: The UARTs definitely need some sort of timing exception because of the combinatorial line-loopback path from RX to TX. I used a multicycle path instead of a false path because if doesn't disable as many design checks.

I'll run a trial to check what happens when I remove the multipath constraints from the non-UART interfaces. Might do a rebase to include pinmux first to avoid later surprises.

elliotb-lowrisc:
Rebased to include pinmux. Reworked clock creation/renaming to use clock net names instead of PLL pins.

elliotb-lowrisc:
Realised the method of finding the PLL pins was faulty, so changed it. Also rebased

elliotb-lowrisc:
Missing line-wrapping `\`'s

data/synth_timing.xdc
GregAC: Thanks for the thorough analysis @elliotb-lowrisc hadn't realised we were looking so tight on the I/O timing. Given this analysis I concur with you the multi-cycle paths are worth keeping, though clearly we need to exercise caution around any RTL that involves them. I wonder if we can do something in simulation with assertions that fire if the relevant signals change more frequently than every 2 clocks?

GregAC: Oh and we are hitting utilization limits which may explain the routing and timing issues. Various things in flight to bring that down significantly.

GregAC (approved):
LGTM thanks for all your work on this @elliotb-lowrisc

elliotb-lowrisc:
Relax some margins to avoid internal setup timing failure on clk_sys

PR #138, opened 2024-07-03
Fixed interrupt indices

HU90m:
After creating https://github.com/lowRISC/sonata-system/issues/137 , realised the comments wouldn't take long to fix, so fixed them.

*Vim's <kbd>Ctrl</kbd>+<kbd>A</kbd> is a god send I reach for a surprising amount.*

PR #190, opened 2024-09-18
Sonata plic - DO NOT MERGE

GregAC:


GregAC:
I can re-generate and rebase this once https://github.com/lowRISC/sonata-system/pull/188 has gone in (adding USBDEV interrupts) so let's not merge this until https://github.com/lowRISC/sonata-system/pull/188 goes in.

GregAC:
Ok rebased on top of the USB interrupt additions. Also thought I'd fixed the licensing lint issues but seems I've actually generated yet more!

GregAC:
Issue was with the lychee link checker, hopefully now resolved

rtl/system/sonata_system.sv
```diff
@@ -319,112 +319,109 @@ module sonata_system #(
...
-    intr_vector[7 +: 1]  = uart0_rx_timeout_irq;
-    intr_vector[6 +: 1]  = uart0_rx_break_err_irq;
-    intr_vector[5 +: 1]  = uart0_rx_frame_err_irq;
-    intr_vector[4 +: 1]  = uart0_rx_overflow_irq;
-    intr_vector[3 +: 1]  = uart0_tx_empty_irq;
-    intr_vector[2 +: 1]  = uart0_rx_watermark_irq;
-    intr_vector[1 +: 1]  = uart0_tx_watermark_irq;
-
-    intr_vector[0 +: 1]  = 1'b0; // This is a special case and tied to zero.
-  end : interrupt_vector
```
HU90m, line 427: `always_comb` vs `assign` was discussed here: https://github.com/lowRISC/sonata-system/pull/138#issuecomment-2208630782

HU90m (commented):
Great to have a sonata specific PLIC!

I'm not keen on all this opentitan tailored documentation making it's way into the repo. It's messy and could cause confusion. It would be nice to create a commented script that generates the PLIC, then removes unnecessary files and patches/removes documentation files. We'd then point to this script in the sonata-system documentation.

GregAC:
> It would be nice to create a commented script that generates the PLIC, then removes unnecessary files and patches/removes documentation files. We'd then point to this script in the sonata-system documentation.

Ah forgot to explicitly discuss this though I do address this in the first commit message. The intent to have auto-generation in but that requires vendoring in ipgen and regen as well as patching them and we may want to update to the latest opentitan whilst we're at it which is a reasonable chunk of work. I want this is ASAP so we can take advantage of the area reduction without being blocked on the larger piece of work.

The auto-generation stuff also interacts with the on-going pinmux work (as it also includes a system top builder which will need to rebuild the PLIC).

rtl/system/sonata_system.sv
GregAC: Ah I missed that I strongly prefer the `assign` style. My feeling is `always_comb` should be used where you actually want to write some logic, here we're just wiring things together.

GregAC: I would also reference our system verilog style guide see:

https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#language-features-1

Where it states 'Prefer assign statements wherever practical.'

HU90m (approved):

alees24:
The 'single IRQ line per block changes' are now in, and you can just (re-)run sw/cheri/tests/test_runner's `plic_tests` to check that all is fine. That should now work reliably for all blocks and multiple iterations; I've made a number of small fixes to the test.

GregAC:
I have accidentally already push this to main! (Admin settings now altered to prevent that happening again...)

However still worth reviewing so updated this PR to allow that

PR #192, opened 2024-09-19
Update PWM documentation to match RTL

elliotb-lowrisc:
There has been a significant divergence between the RTL and the documentation.
Update the documentation to match the RTL to avoid changing the (working) design at this point.


elliotb-lowrisc:
Ah, should be 12-channel 8-bit, not the other way around

marnovandermaas (approved):
Thanks for changing this. Since the current documentation is inaccurate lets merge this and we can always fix it with follow up PRs.

PR #198, opened 2024-09-19
Migrate SPI CS control into the controllers

alees24:
This PR collects a few SPI-related changes:

1. Modify the SPI controller to incorporate a parameterised number of CS (Chip Select) lines to the attached peripherals.
(Since the SPI controllers share a single auto-generated register implementation the intention is to set the parameter to be large enough that all SPI controllers are accommodated, and the surplus CS lines in some controls will simply be unused and unconnected.
2. Reduce the size of the SPI FIFOs to 16 entries.
3. Report the size of the SPI FIFOs in the new 'INFO' register so that driver software may adapt*.
4. Implement the interrupt logic in the SPI controller.
5. Update software to  use the new CS register.

**Note** the software changes depend upon an update to platform-spi.hh in the CHERIoT-RTOS repository (# 4).

* it would be worth presenting to software the number of entries available in the 'tx' fifo rather than the number of used entries; this is both more useful and avoids the issue of driver software needing to know the fifo size.


HU90m (commented):
The chip select (CS) register layout looks sensible to me.

If we want auto chip select in the future, we could add a second register which like the CS register has a bit for each CS line. When this bit is set high, the auto chip select is enabled and the CS register becomes read-only.

dv/verilator/top_verilator.sv
```diff
@@ -157,6 +157,7 @@ module top_verilator (input logic clk_i, rst_ni);
 
   logic spi_rx[SPI_NUM];
   logic spi_tx[SPI_NUM];
+  logic [SPI_CS_NUM-1:0] spi_cs[SPI_NUM];
```
GregAC, line 160: nit: align `spi_rx`/`spi_tx`/`spi_clk` with `spi_cs`

GregAC (approved):
LGTM modulo one nit, thanks @alees24 

We should write some test to check the interrupts actually work, but no required for this PR. Maybe time to spin up an initial block level DV environment for this...

rtl/ip/spi/rtl/spi.sv
```diff
@@ -3,24 +3,29 @@
 // SPDX-License-Identifier: Apache-2.0
 
 module spi import spi_reg_pkg::*; #(
-  parameter int unsigned RxFifoDepth = 64,
-  parameter int unsigned TxFifoDepth = 64
+  parameter int unsigned RxFifoDepth = 16,
```
GregAC, line 6: FYI at the last sunburst meeting we decided we should go for a depth of 8 for FIFOs across the board. The thinking being right now tiny FIFOs are fine and it's easier to increase the depth rather than decrease.

alees24:
> LGTM modulo one nit, thanks @alees24
> 
> We should write some test to check the interrupts actually work, but no required for this PR. Maybe time to spin up an initial block level DV environment for this...

Writing an interrupt test is on my ToDo list.

marnovandermaas (changes requested):
Adding this request for change because CI lint is failing.

rtl/ip/spi/rtl/spi.sv
```diff
@@ -166,11 +207,77 @@ module spi import spi_reg_pkg::*; #(
     .spi_clk_o
   );
 
-  assign hw2reg.intr_state = '0;
+  // CS signals come directly from the register interface and shall be modified only when the
+  // core is Idle.
```
marnovandermaas, line 211: What is stopping this update from happening when the block isn't idle?

rtl/ip/spi/rtl/spi_reg_top.sv
```diff
@@ -1,4 +1,4 @@
-// Copyright lowRISC contributors.
+// Copyright lowRISC contributors (OpenTitan project).
```
marnovandermaas, line 1: I know this is autogenerated but not sure we want to add this.

rtl/system/sonata_pkg.sv
```diff
@@ -11,4 +11,6 @@ package sonata_pkg;
   localparam int unsigned I2C_NUM  = 2;
   localparam int unsigned SPI_NUM  = 5;
 
+  localparam int unsigned SPI_CS_NUM = 4;
```
marnovandermaas, line 14: This file is auto generated did you add it in the template?

rtl/system/sonata_system.sv
```diff
@@ -5,7 +5,9 @@
 // The Sonata system, which instantiates a CHERIoT Ibex, TileLink Uncached
 // Lightweight bus and a number of common peripherals, usc as I2C, SPI, UART,
 // USB.
-module sonata_system #(
+module sonata_system
+  import sonata_pkg::*;
```
marnovandermaas, line 9: Nice! Didn't know you could do this :)

marnovandermaas (commented):
Some comments from just reading the code.

rtl/ip/spi/rtl/spi.sv
alees24: We are relying upon software to modify the CS lines only when the block is idle, just as when it was using the GPIO lines; it can ascertain the idle/busy state of the SPI core from the `status` register.

rtl/system/sonata_pkg.sv
alees24: Thanks for the prompt; I hadn't registered that it was generated.

rtl/ip/spi/rtl/spi.sv
GregAC: I happy to leave this behaviour as is given it's exactly as it was before (i.e. SW needs to know when it's safe to alter CS line states).

We could always add this as a new feature. Some argument as it's software visible we should do this pre v1.0 but feels like an innocuous enough change to me (shades of https://xkcd.com/1172/ for anyone who complains being able to alter CS whilst the block is active is an essential feature ;) ).

GregAC:
@alees24 would be good to get this merged. Looks like the current CI failure is just down to https://github.com/lowRISC/cheriot-rtos/pull/4 needing to be merged so we can update the CHERIoT RTOS commit used by cmake for the sonata-system software build.

alees24:
Thanks to @hcallahan-lowrisc and @HU90m for help sorting out the cross-repo fun.

marnovandermaas (approved):
Thanks lets merge this.

PR #200, opened 2024-09-20
Add SPI test for CHERI

AlexJones0:
This PR adds a SPI test to the existing baremetal CHERI test runner. For now, this is a simple smoketest that simply tests a few commands for the SPI flash - other SPI connections are not included.

The SPI tests consist of 3 components:
1. Read the JEDEC ID from flash over SPI and check it matches the value we expect.
2. Erase 1 random sector (4 KiB) from flash over SPI, and check that the entire sector was properly erased.
3. Write 10 random pages (256 B) to flash over SPI, and then read back the data to check that it matches what was written.

The number of test iterations and the number of sectors and pages can each be configured via compilation flags. These numbers are purposefully kept fairly low to keep runtime reasonable on Verilator.

This commit also includes minor changes to the `flash-utils.hh` simple SPI driver, to allow it to use 4-byte addresses required for testing the entirety of the flash via SPI.

This has been tested on both FPGA and in Verilator, and has also been confirmed to pass 100 iterations with 10 sector erases and 50 page writes each for consistency on FPGA.

### DIFF
--- a/sw/cheri/common/flash-utils.hh
+++ b/sw/cheri/common/flash-utils.hh
@@ -19,8 +19,10 @@ static const uint8_t CmdReset               = 0x99;
 static const uint8_t CmdReadJEDECId         = 0x9f;
 static const uint8_t CmdWriteEnable         = 0x06;
 static const uint8_t CmdSectorErase         = 0x20;
+static const uint8_t CmdSectorErase4Addr    = 0x21;
 static const uint8_t CmdReadStatusRegister1 = 0x05;
 static const uint8_t CmdPageProgram         = 0x02;
+static const uint8_t CmdPageProgram4Addr    = 0x12;
 static const uint8_t CmdReadData            = 0x03;
 static const uint8_t CmdReadData4Addr       = 0x13;
 
@@ -56,15 +58,15 @@ class SpiFlash {
   }
 
   void erase_sector(uint32_t address) {
-    const uint8_t erase_cmd[4] = {CmdSectorErase, uint8_t((address >> 16) & 0xff), uint8_t((address >> 8) & 0xff),
-                                  uint8_t(address & 0xff)};
+    const uint8_t erase_cmd[5] = {CmdSectorErase4Addr, uint8_t((address >> 24) & 0xff), uint8_t((address >> 16) & 0xff),
+                                  uint8_t((address >> 8) & 0xff), uint8_t(address & 0xff)};
 
     set_cs(true);
     spi->blocking_write(&CmdWriteEnable, 1);
     set_cs(false);
 
     set_cs(true);
-    spi->blocking_write(erase_cmd, 4);
+    spi->blocking_write(erase_cmd, 5);
     set_cs(false);
 
     set_cs(true);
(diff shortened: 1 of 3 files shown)


AlexJones0:
After seeing the issues in `flash-utils.hh` found by #201, I've made the tests slightly more rigorous so that they ensure that writes/reads are actually happening and not just passing because of deterministic PRNG over test iterations. I've also added an additional fourth test that slows down the SPI clock 8 times and tests writing random data to a random page, to try and catch the issue described there. As suggested in that PR, this new fourth test does indeed fail at the moment as you would expect. All three other tests still pass consistently, as previously described.

AlexJones0:
I've also now introduced a commit based on PR #201 that fixes the failing slow SPI clock speed test now. Unlike that PR, this retains the CS falling/rising edges so that instructions are properly executed, and it uses appropriate `wait_idle` calls to ensure that relevant FIFOs are empty before the CS is modified. I've verified that now with that commit, all four SPI tests pass on both the Verilator and FPGA environment as described before.

sw/cheri/tests/spi_tests.hh
```diff
@@ -0,0 +1,303 @@
...
+static constexpr unsigned SpiFlashPageSize   = 256;               // 256 B
+static constexpr unsigned SpiFlashSectorSize = 4 * 1024;          // 4 KiB
+static constexpr unsigned SpiFlashSize       = 32 * 1024 * 1024;  // 32 MiB
+static constexpr size_t SpiFlashPages        = SpiFlashSize / SpiFlashPageSize;
+static constexpr size_t SpiFlashSectors      = SpiFlashSize / SpiFlashSectorSize;
+
+/**
+ * Get the start address of a random page in flash.
+ */
+__always_inline uint32_t random_flash_page_addr(ds::xoroshiro::P32R8 &prng) {
```
HU90m, line 62:
```suggestion
[[always_inline]] uint32_t random_flash_page_addr(ds::xoroshiro::P32R8 &prng) {
```

Also why `always_inline`? Is it important that it's inlined?

sw/cheri/tests/spi_tests.hh
```diff
@@ -0,0 +1,303 @@
...
+ * the SPI has already been appropriately configured. The address is
+ * assumed to be sector-aligned.
+ * Returns the number of failures during the test.
+ */
+static int spi_flash_erase_test_sector(Capability<volatile SonataSpi> spi, ds::xoroshiro::P32R8 &prng,
+                                       SpiFlash spi_flash, uint32_t start_addr) {
+  int failures = 0;
+  std::array<uint8_t, SpiFlashPageSize> write_values, read_values;
+
+  // Write some random values to the first page of the specfied sector
```
HU90m, line 90:
```suggestion
  // Write some value to the first page of the specified sector
```

HU90m (approved):

sw/cheri/tests/spi_tests.hh
AlexJones0: No - I think I just included this out of habit because it's a small piece of logic that I extracted out of two functions to reduce repetition. I'm happy to get rid of this.

sw/cheri/common/flash-utils.hh
```diff
@@ -38,10 +40,12 @@ class SpiFlash {
   void reset() {
     set_cs(true);
     spi->blocking_write(&CmdEnableReset, 1);
+    spi->wait_idle();
```
nbdd0121, line 43: In long term, when CS is managed by SPI host, we probably want to have `set_cs(false)` automatically execute `wait_idle()`

PR #201, opened 2024-09-20
Do not modify CS whilst SPI controller is active

alees24:
Ensure that CS is not changed whilst the SPI controller is active, e.g. after a write into the FIFO that has not necessarily completed on the SPI bus.
It is safe to deassert the CS line when a read operation has completed because that necessarily blocks until completion.

The code worked in practice by dint of the CPU writes into the FIFO being slow enough that the SPI at 15Mbps can mostly keep up. If the SPI is slowed to eg. 30/8Mbps then the code `sw/cheri/checks/spi_test` produced incorrect results in Verilator simulation because the spidpi model responds to the changes on the CS line.

Such issues with the CS line-driving may exist elsewhere in the codebase and particularly for other SPI devices.

### DIFF
--- a/sw/cheri/common/flash-utils.hh
+++ b/sw/cheri/common/flash-utils.hh
@@ -46,10 +46,10 @@ class SpiFlash
 	{
 		set_cs(true);
 		spi->blocking_write(&CmdEnableReset, 1);
-		set_cs(false);
 
-		set_cs(true);
 		spi->blocking_write(&CmdReset, 1);
+
+		spi->wait_idle();
 		set_cs(false);
 
 		// Need to wait at least 30us for the reset to complete.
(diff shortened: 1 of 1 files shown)


engdoreis (approved):

sw/cheri/common/flash-utils.hh
```diff
@@ -46,10 +46,10 @@ class SpiFlash
 	{
 		set_cs(true);
 		spi->blocking_write(&CmdEnableReset, 1);
-		set_cs(false);
```
nbdd0121, line 49: The winbond datasheet says that instructions should be initiated with the falling edge of C̅S̅ and completed with rising edge of it.

HU90m:
Closing issue should be fixed by the added `wait_idle`s in https://github.com/lowRISC/sonata-system/pull/200

PR #202, opened 2024-09-23
Add I2C test for CHERI

AlexJones0:
This PR adds an I2C test to the existing baremetal CHERI test runner. For now, these are simple smoketests that test a few possible connected I2C devices by reading out some information and checking it matches what we expect. The tests are made based on the existing `i2c_hat_id` demo being used as a non-CHERI check (see [`i2c_hat_id.c`](https://github.com/lowRISC/sonata-system/blob/main/sw/legacy/demo/i2c_hat_id/i2c_hat_id.c)), and have been adapted for CHERI. Please suggest if you have any other ideas for small tests that you think would be appropriate to add here.

There are three tests, which can be configured based on which devices are present:
  1. Read the Raspberry Pi HAT ID EEPROM header and check it.
  2. Read the Raspberry Pi HAT Accelerometer & Gyroscope and Magnetic Sensor `WHO_AM_I` registers and check these.
  3. Read the config and temperature value from the AS6212 Temperature sensor and check these values are sensible.

The number of test iterations, and the devices that are available, as well as whether to check the contents of the RPi HAT ID EEPROM header, can all be configured via compilation flags.

This has been tested on both FPGA and in Verilator. Currently, it passes on FPGA with the devices connected and fails on Verilator. With #194 applied on top, it passes on Verilator also. This test has also been confirmed to pass 100 iterations for consistency on FPGA.


sw/cheri/tests/i2c_tests.hh
```diff
@@ -0,0 +1,285 @@
...
+  // Send two 0x0000 byte addresses and skip the STOP condition.
+  const uint8_t addr[] = {0, 0};
+  i2c->blocking_write(RpiHatIdEepromAddr, addr, sizeof(addr), true);
+
+  // Fill a read buffer with dummy data so that we see changes.
+  constexpr uint8_t DummyReadVal = 0x3D;
+  uint8_t eeprom_data[RpiHatIdEepromHeaderSize];
+  for (auto &read_val : eeprom_data) {
+    read_val = DummyReadVal;
+  }
```
engdoreis, line 138:
```suggestion
  std::array<uint8_t,RpiHatIdEepromHeaderSize> eeprom_data;
  eeprom_data.fill(DummyReadVal);
```

sw/cheri/tests/i2c_tests.hh
```diff
@@ -0,0 +1,285 @@
...
+
+  // Read a page from the RPI Hat ID EEPROM
+  if (!i2c->blocking_read(RpiHatIdEepromAddr, eeprom_data,
+                          RpiHatIdEepromHeaderSize)) {
+    failures++;
+  }
+
+  // Verify the contents of the EEPROM header that were read.
+  bool buffer_changed = false;
+  for (uint32_t index = 0; index < RpiHatIdEepromHeaderSize; index++) {
```
engdoreis, line 148: Use iterator?

AlexJones0: Keeping this because I'm simultaneously iterating over two arrays. I could perhaps get an iterator for both arrays and iterate through each in a loop, but for simplicity I'm just keeping a simple index-based loop here.

sw/cheri/tests/i2c_tests.hh
```diff
@@ -0,0 +1,285 @@
...
+constexpr uint8_t As6212Addr = 0x48;
+
+/**
+ * The addresses of various I2C device registers used in testing.
+ */
+constexpr uint8_t RpiHatWhoAmIRegAddr = 0x0F;
+constexpr uint8_t As6212TempRegAddr = 0x00;
+constexpr uint8_t As6212ConfigRegAddr = 0x01;
+
+bool i2c_interrupt_flag = false;
```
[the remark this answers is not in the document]
AlexJones0: Oops, leftover from previous testing. Good catch!

sw/cheri/tests/i2c_tests.hh
AlexJones0: I was doing this because `memset` is not available. I've opted to use this suggested approach instead, and changed the reference to `eeprom_data` to `eeprom_data.data()` as the I2C read function expects a C-style array.

sw/cheri/tests/i2c_tests.hh
```diff
@@ -0,0 +1,267 @@
...
+  }
+  return failures;
+}
+
+/**
+ * Run the whole suite of I2C tests.
+ */
+void i2c_tests(CapRoot root, UartPtr console) {
+  // Retrieve bounded capabilities for both I2C controllers
+  I2cPtr i2c0 = i2c_ptr(root, 0);
```
engdoreis, line 239: I would move these declarations closer to where it's used.

engdoreis (approved):
LGTM.

AlexJones0:
Test is failing CI because #194 has not yet been merged.

engdoreis (approved):

HU90m (approved):

PR #204, opened 2024-09-24
Introduce disable_cheri runtime switch

HU90m:
Simulation builds default to CHERI, but legacy software may be executed on the same build by supplying the command-line option `+disable_cheri` when running the simulator.

Supersedes #191


dv/verilator/top_verilator.sv
```diff
@@ -106,11 +105,18 @@ module top_verilator (input logic clk_i, rst_ni);
   // Reporting of CHERI enable/disable and any exceptions that occur.
   wire  [CheriErrWidth-1:0] cheri_err;
   logic [CheriErrWidth-1:0] cheri_errored;
-  wire cheri_en;
-  initial begin
-    if (cheri_en) $display("Running with CHERI enabled");
-    else $display("Running in legacy software mode");
-  end
+  logic cheri_en_i;
```
rswarbrick, line 108: I'd probably drop the "_i" from the name variable here because it's not an input to `top_verilator`.

rswarbrick (commented):
Other than the variable name question, this looks perfectly sensible and I'm keen!

rswarbrick (approved):
Looks great, thanks!

PR #210, opened 2024-09-26
Add setuptools to the python dependencies

HU90m:
The `pkg_resources` module is required by
`vendor/lowrisc_ip/util/tlgen/generate.py`.


AlexJones0 (approved):

nbdd0121:
Eh. After https://github.com/lowRISC/opentitan/pull/24605 is merged, we should revendor and undo this change.

PR #212, opened 2024-09-27
Add simple Pinmux driver & check application.

AlexJones0:
This PR adds a simple Pinmux driver, and a check application to go along with it.
More detailed information can be found in commit messages.

The Pinmux driver is a slightly modified version of one that I've already been using to develop applications running on CHERIoT RTOS, with small modifications such that it takes a bounded capability as input rather than using the RTOS `MMIO_CAPABILITY` macro recommended to be used in the RTOS drivers. For now this driver requires users to read the documentation to see what is muxed where; it might be nice to extend this to enforce compile-time checks in the future but this was proving sufficiently complicated for now, and more emphasis is placed on testing the hardware.

The intent is that the driver exists in `sonata-system` for usage in check and CI test applications (similar to the USB driver), and so that it can live somewhere while there are potential software-breaking changes to the pinmux RTL that should be tested and fixed before the driver is upstreamed.

The PR also adds a `pinmux_check` check application, which simply logs some messages to the console UART (UART0), whilst enabling and disabling the output of UART0 to ser0_tx via pinmux, such that some log messages are lost and not seen. A user can then read the messages to confirm that the pinmux appears to be working at some basic level. Testing with this basic check was sufficient to notice the RTL error in #209, for example.

This can be tested by running the `pinmux_check` application directly and confirming that you get the expected result.
Note that this currently is only confirmed to work on FPGA and not on Verilator, as pinmux is not working in the Verilator simulation environment at the time of writing.

**Edit**: The driver in this PR is now also being used in more software/tests by #227 and #230, and I haven't encountered any issues using it to mux pins for GPIO, UART, I2C or SPI; you can also use those PRs to test this PR.


sw/cheri/common/platform-pinmux.hh
```diff
@@ -0,0 +1,360 @@
...
+/**
+ * A driver for Sonata's Pin Multiplexer (Pinmux).
+ *
+ * This driver can be used to select which block output is placed on a given
+ * output pin, and to select which input pin is provided to a given block
+ * input. Sonata's pinmux only allows certain selections per output pin / block
+ * input, which this driver describes.
+ *
+ * Documentation source can be found at:
+ * https://github.com/lowRISC/sonata-system/blob/6914b30a2769f881a53afb912032cd65a6d14307/doc/ip/pinmux.md
```
GregAC, line 21: I think better not to link to a specific commit here, could make this an in tree reference (Documentation source can be found in the repository at `doc/ip/pinmux.md`) and then leave the rendered doc link as is?

sw/cheri/common/platform-pinmux.hh
```diff
@@ -0,0 +1,360 @@
...
+   * multiplexed, meaning that they can be changed to output the outputs
+   * of different blocks (or disabled). The block outputs that can be
+   * selected are limited, and vary on a per-pin basis.
+   *
+   * Documentation sources:
+   * https://lowrisc.github.io/sonata-system/doc/ip/pinmux.html
+   * https://github.com/lowRISC/sonata-system/blob/4b72d8c07c727846c6ccb27754352388f3b2ac9a/data/pins_sonata.xdc
+   * https://github.com/newaetech/sonata-pcb/blob/649b11c2fb758f798966605a07a8b6b68dd434e9/sonata-schematics-r09.pdf
+   */
+  enum class OutputPin : uint16_t {
```
GregAC, line 56: Don't need to do this now but we should look at auto-generating these from the pinmux configuration

GregAC (approved):

sw/cheri/common/platform-pinmux.hh
AlexJones0: I used a specific commit because that is what the existing drivers in `cheriot-rtos` do, to ensure that they point to the documentation at the time that the driver was written (which makes it easy to identify if a driver is bugged or has just fallen out of date). Because this driver is only intended to live in `sonata-system` temporarily whilst we change and iterate on the pinmux config (avoid breakages from PRs across multiple repos), it doesn't make sense to make an in-tree reference when this will eventually exist at `cheriot-rtos/sdk/include/platform/sunburst/platform-pinmux.hh`.

sw/cheri/common/platform-pinmux.hh
AlexJones0: I was considering this whilst doing it. At the very least it would be nice to have some matching names in the auto-generated documentation so that users can more easily identify what is muxable where. This would have to be a separate script though I think, as this driver is intended to eventually live in upstream `cheriot-rtos` and not in this repo.

sw/cheri/common/platform-pinmux.hh
```diff
@@ -0,0 +1,360 @@
...
+    return true;
+  }
+
+  /**
+   * A constructor for the SonataPinmux driver, which takes a bounded
+   * capability to the pinmux registers. This should be replaced with
+   * an appropriate `MMIO_CAPABILITY` call in the version of the driver
+   * that runs in CHERIoT RTOS, rather than baremetal.
+   */
+  SonataPinmux(volatile uint8_t *registers) { this->registers = registers; }
```
engdoreis, line 359: Nit. 
```suggestion
  SonataPinmux(volatile uint8_t *registers) : registers(registers) { }
```

engdoreis (commented):
This PR looks good to me, however I think the last commit should be squashed into the first before merging.

AlexJones0:
Push has rebased on main, addressed the nit and squashed the first and last commit.

PR #213, opened 2024-09-28
Re-add CHERIoT RTOS test suite to Nix testing infrastructure.

AlexJones0:
See commits for more detailed explanations.

This PR re-adds testing of the CHERIoT RTOS test suite in our test suite, such that it will now also run in CI where appropriate. To make the test suite work, the CHERIoT RTOS source used in `flake.nix` has been bumped to point to the [`hyperram` branch](https://github.com/lowRISC/cheriot-rtos/tree/hyperram), instead of `main`. To keep these in sync, the corresponding git tag in the relevant CMake build file is also updated. With these changes, the test suite should now once again pass as expected. 

It also doubles the time available for regular tests to run in simulation in CI, because we're getting quite close to the timeout on longer variations after adding the SPI, I2C and hyperram tests, and it already times out locally on my machine.

Not got a board with me at the time of writing, so I haven't had a chance to check this out on FPGA yet - only on simulator.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -161,9 +161,9 @@
       cheriotRtosSource = pkgs.fetchFromGitHub {
         owner = "lowRISC";
         repo = "CHERIoT-RTOS";
-        rev = "30148bad7ef45026d4e4b754871bf2b396ce2ca8";
+        rev = "444e4965c1dc84414e4959d55e47b0e999c086c9";
         fetchSubmodules = true;
-        hash = "sha256-3yatAEfZ5eWOfQB7w5I6OOHaG39pRUeyzCGgix3T4PI=";
+        hash = "sha256-tXTIKmf/MBXmspcPdKYqQQp8zj6HuUgGXlj+uUCcC40=";
       };
 
       sonata-system-software = pkgs.stdenv.mkDerivation rec {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -161,9 +161,9 @@
       cheriotRtosSource = pkgs.fetchFromGitHub {
         owner = "lowRISC";
         repo = "CHERIoT-RTOS";
-        rev = "30148bad7ef45026d4e4b754871bf2b396ce2ca8";
+        rev = "444e4965c1dc84414e4959d55e47b0e999c086c9";
```
HU90m, line 164: Can you sync this up with the version here: https://github.com/lowRISC/sonata-system/blob/cae088cf016aa3b26fc3b3dfbb2623282e4fdb8d/sw/cheri/CMakeLists.txt#L10

HU90m (changes requested):
Brilliant! Thanks Alex

flake.nix
AlexJones0: If I don't point to the `hyperram` branch as described then the RTOS cannot be built & run in CI - the current Git tag being used in the `CMakeLists.txt` file does not include the hyperram hack.
I've instead updated the commit to sync the CMake CHERIoT RTOS soruce up with this nix change.

HU90m (approved):

PR #215, opened 2024-10-01
Changed verilator's scope to top_verilator

HU90m:
This fixes the problem of `mhpmcounter_get` not being in scope.


alees24 (approved):
brillianT work well done!

PR #216, opened 2024-10-03
GPIO output enable now readable

marnovandermaas:
@AlexJones0 can you test whether this works for you?


AlexJones0 (approved):
From testing in software this seems to work. Both from simply writing/reading all 1s and using it to read and set specific GPIO pins as input/output appears to work as expected.

I would recommend that someone who can better read the RTL reviews this before merging, but from the software side of things it seems to be all good.

PR #218, opened 2024-10-03
SPI/LCD check code

alees24:
Introduce lcd_check which drives the SPI controller and LCD at maximum speed to exercise/test the reliability of the core/FIFO logic.
Support transposed rendering and output in the SPI LCD model.
~~This depends on PR #198~~


sw/cheri/checks/lowrisc_logo_native.h
marnovandermaas, line 1: How did you generate this?

marnovandermaas (commented):
Looks reasonable to me. Especially in the checks folder it is nice to have something like this that keeps running. We should fix CI before merging though.

sw/cheri/checks/lowrisc_logo_native.h
alees24: Not exactly; it's just converted into the native LCD format from the file `sw/legacy/demo/lcd_st7735/lowrisc_logo.h` I've added a comment to that effect.

AlexJones0:
@alees24 Here is the fix to Nix to get this working in CI:

```patch
diff --git a/nix/software.nix b/nix/software.nix
index 9224241e..bbaa45ab 100644
--- a/nix/software.nix
+++ b/nix/software.nix
@@ -19,13 +19,14 @@ in {
   sonata-system-software = pkgs.stdenv.mkDerivation rec {
     name = "sonata-system-software";
     src = fileset.toSource {
-      root = ../sw;
+      root = ../.;
       fileset = fileset.unions [
         ../sw/cheri
         ../sw/common
+        ../vendor/display_drivers/st7735
       ];
     };
-    sourceRoot = "${src.name}/cheri";
+    sourceRoot = "${src.name}/sw/cheri";
     nativeBuildInputs = cheriotPkgs ++ (with pkgs; [cmake]);
     cmakeFlags = [
       "-DFETCHCONTENT_SOURCE_DIR_CHERIOT_RTOS=${cheriotRtosSource}"
(nix:sonata-system-devshell-env) [15:56:14] [alexj] [~/lr/sonata-syst
```

alees24:
> @alees24 Here is the fix to Nix to get this working in CI:
> ...

Thanks for that help.

sw/cheri/checks/lowrisc_logo_native.h
marnovandermaas: Thanks!

marnovandermaas (approved):
Tested this after a rebase on main so I think this is ready to merge.

dv/dpi/spidpi/spi_lcd.hh
```diff
@@ -45,6 +46,8 @@ private:
 
   bool pixWriting;  // Pixel writing mode?
 
+  uint8_t  madCtl;  // Memory Data Access Control.
```
HU90m, line 49: Should this be `mdaCtrl`?

PR #220, opened 2024-10-07
Rework GPIO debounce to save area

elliotb-lowrisc:
Use a shared debounce counter for all pins, rather than a counter each.

### DIFF
--- a/rtl/system/debounce_step.sv
+++ b/rtl/system/debounce_step.sv
@@ -1,0 +1,44 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Maintain a 1-bit 'counter' that increments when a step-pulse is received
+// and the input is in the opposite state from the debounced output.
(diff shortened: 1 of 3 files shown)


elliotb-lowrisc:
Reduces `u_gpio` resource usage to around 410 LUTs & 904 Flops according to `report_utilization -hierarchical`.

GregAC:
> Reduces u_gpio resource usage to around 410 LUTs & 904 Flops 

Was hoping we'd get it down yet more but I guess we need a handful of flops per GPIO due to the sync flops as well as the output/output enable and debounce stuff. I guess we could only put debounce on a subset of the GPIOs but probably won't save us that much.

HU90m (changes requested):
This looks great thanks Eliot!

Can we just replace the existing `debounce` module, i.e. remove it and rename `debounce_step`?

marnovandermaas:
> This looks great thanks Eliot!
> 
> Can we just replace the existing `debounce` module, i.e. remove it and rename `debounce_step`?

I looked at this but debounce is still used in the reset controller.

HU90m (approved):

Issue #6, opened 2023-12-01
FPGA versioning


marnovandermaas:
Add a versioning mechanism for software to detect which FPGA build is currently configured. This is to be used as a debugging mechanism so that software does not lock up when it is running on a different FPGA build.

We can store a hash of the FPGA configuration in a CSR or a memory mapped register to compare to.

PR #221, opened 2024-10-07
System info IP

marnovandermaas:
This adds an IP block called system info, which currently contains the git hash of the current head and the dirty status of the repo. These values get automatically populated when running FuseSoC.

Closes: https://github.com/lowRISC/sonata-system/issues/6


GregAC (changes requested):
Will review in more detail later but putting on request changes as I'd like to stop the proliferation of manually written register interfaces. Effectively they're a legacy thing due to starting with ibex demo system.

Going forward we should use reggen in all cases unless there's specific practical difficulties in doing so (e.g. as we have seen in the XADC block).

Please also put this in `'rtl/ip' under it's own 'system_info' directory (matching the structure of other IP directories) which will include the reggen hjson.

HU90m (dismissed):
I think this may make more sense as a FuseSoC generator called at build time. Then we can be pretty certain the info accurate for a given build. We're not left wondering whether someone had forgot to re-run `top_gen.py`.

marnovandermaas:
Converting to draft until I've addressed automatic generation via FuseSoC and reggen.

marnovandermaas:
Ok, this PR is now ready for review. It took quite some time to do both of @HU90m and @GregAC comments, but I think this is all in there now.

rtl/ip/system_info/data/system_info.hjson
```diff
@@ -0,0 +1,61 @@
...
+      fields: [
+        { bits: "31:0",
+          name: "HASH",
+          desc: '''First 32-bits of the SHA hash.'''
+        }
+      ]
+    },
+    { name: "RTL_COMMIT_DIRTY",
+      desc: '''Bytes written here are pushed to the transmit FIFO. If the FIFO
+               is full writes are ignored.''',
```
elliotb-lowrisc, line 48: ?

rtl/ip/system_info/data/system_info.hjson
```diff
@@ -0,0 +1,61 @@
...
+  available_output_list: [],
+  interrupt_list: [],
+  regwidth: "32",
+  registers: [
+    { name: "RTL_COMMIT_HASH",
+      desc: '''This contains the git hash of the Sonata System repository when this bitstream was generated.
+               This is useful to determine the configuration of the RTL.''',
+      swaccess: "ro",
+      hwaccess: "hwo",
+      hwext: "true",
```
elliotb-lowrisc, line 38: May want to use this with `RTL_COMMIT_DIRTY` too to remove the `de` bit from `system_info_hw2reg_rtl_commit_dirty_reg_t`

elliotb-lowrisc (changes requested):

rtl/ip/system_info/data/system_info.hjson
elliotb-lowrisc: I think not setting the `de` bit is actually causing `system_info_check` to always report "clean". I can't get it to report "DIRTY!" even when I see it trying to set that bit in the generated system_info.sv file.

rtl/ip/system_info/data/system_info.hjson
marnovandermaas: Thanks fixed!

rtl/ip/system_info/data/system_info.hjson
marnovandermaas: Good point. I had tested the commit part quite thoroughly but not the dirty bit. I also took the liberty to make the dirty bit reset to 1, not 0.

elliotb-lowrisc (approved):
Works both dirty and clean for me now

rtl/ip/system_info/util/system_info_gen.py
```diff
@@ -0,0 +1,71 @@
...
+    try:
+        repo = git.Repo(search_parent_directories=True)
+        commit = repo.head.object.hexsha
+        dirty = repo.is_dirty()
+    except git.exc.InvalidGitRepositoryError:
+        commit = os.environ.get("FLAKE_GIT_COMMIT")
+        dirty = bool(os.environ.get("FLAKE_GIT_DIRTY"))
+        if commit is None or dirty is None:
+            print("Error: Couldn't get commit and dirty via git or Nix")
+            exit()
```
hcallahan-lowrisc, line 29: It's probably a little cleaner to reverse the order of the check (check if the FLAKE_* envvars exist first), and then `import git # noqa: E402` inside the else path. Then you won't need to add pkgs.git dependency for the nix builds (sonata-simulator{-lint}).

rtl/ip/system_info/util/system_info_gen.py
```diff
@@ -0,0 +1,71 @@
...
+        (
+            system_info_dir / "templates" / "system_info.sv.tpl",
+            Path("system_info.sv"),
+        ),
+        (
+            system_info_dir / "templates" / "system_info.core.tpl",
+            Path("system_info.core"),
+        ),
+        # TODO is there a way not to have to copy these over to the
+        # auto-generated folder?
```
hcallahan-lowrisc, line 50: Is this comment now stale?

marnovandermaas:
@alees24 had the idea of adding the system frequency as well as the number of I2C, UART, SPI and GPIO blocks in the system information IP block.

rtl/ip/system_info/util/system_info_gen.py
marnovandermaas: Ah interesting I hadn't thought of doing the input inside the if statement. I do find it weird that it requires the git executable just for importing the library. Should now be fixed.
Ruff check removed the `noqa` comment, I hope that's ok.

marnovandermaas:
Added system frequency and block instances to the system info block. And added @alees24 as a reviewer.

doc/ip/system_info.md
```diff
@@ -0,0 +1,11 @@
+# System information
+
+This is a special IP block that is automatically populated with useful information about the system.
+These values are useful to know what the parameters were when a bitstream was generate.
+
+| Offset | Register |
+|--------|----------|
+| 0x0000 | Git commit hash from which this bistream was generated.
+| 0x0004 | Git dirty status (1 if dirty or 0 if clean).
+| 0x0008 | System clock frequency in Hz.
```
nbdd0121, line 10: Oh no, sonata is not limited to <4.3 GHz in perpetuity!

doc/ip/system_info.md
```diff
@@ -0,0 +1,11 @@
+# System information
+
+This is a special IP block that is automatically populated with useful information about the system.
+These values are useful to know what the parameters were when a bitstream was generate.
+
+| Offset | Register |
+|--------|----------|
+| 0x0000 | Git commit hash from which this bistream was generated.
```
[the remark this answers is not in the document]
GregAC: We could just have the full hash available? Bit of a waste of LUTs I guess but it's not that many....

Though of course people commonly shorten hashes when referring to git objects without issue, e.g. going by https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#Short-SHA-1 in the Linux kernel 12 SHA-1 digits (so 12 * 4 = 48 bits) is enough to uniquely identify all objects. We could expand to 64 bits here? 32-bits almost certainly fine for our use case but 64 certainly feels more comfortable.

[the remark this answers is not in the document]
nbdd0121: Yeah, full hash is too wasteful, 64 bit would be plenty, although I think it's quite unlikely for us to have a collision with 32 bits.

doc/ip/system_info.md
marnovandermaas: I was thinking about this, but didn't find this a good enough reason to switch to kHz ;)

doc/ip/system_info.md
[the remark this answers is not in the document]
marnovandermaas: Ok, this has now been extended.

doc/ip/system_info.md
```diff
@@ -0,0 +1,11 @@
...
+| 0x0004 | Git dirty status (1 if dirty or 0 if clean).
+| 0x0008 | System clock frequency in Hz.
+| 0x000C | IP block instances packed as 4 bytes from most significant byte to least significant byte: SPI, I2C, UART and GPIO.
```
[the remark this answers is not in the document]
marnovandermaas: Good point. I've now split them up into their own info registers. That means we can possibly add extra information in the top 3 bytes in the future.

PR #222, opened 2024-10-08
Rework resets

elliotb-lowrisc:
Add reset deassertion synchronisation for clk_sys and clk_usb.

Rationalise reset structure for clk_hr.

The result is the removal of many difficult reset timing paths and experimentally-observed improvement to reset reliability.


elliotb-lowrisc:
Add missing simulator/unused signals

nbdd0121:
I still think we should have a PoR reset being the only thing in rtl/fpga/rst_ctrl.sv, and have the rest of resets being generated from that one and then we can provide SW interface to the reset the top.

elliotb-lowrisc:
> I still think we should have a PoR reset being the only thing in rtl/fpga/rst_ctrl.sv, and have the rest of resets being generated from that one and then we can provide SW interface to the reset the top.

I don't think I quite understand what you mean by a SW interface to reset the top. Microcontrollers don't usually have a direct way to trigger a reset from software to my knowledge. They have power-on reset, brown-out reset, watchdog timer reset, external reset, etc.

Separately, my preference is for all the resets to be generated in one place, ideally near where the clocks are generated. Makes it easier to discover/maintain, and replace if needed.

nbdd0121:
You can see how opentitan does it here: https://opentitan.org/book/hw/top_earlgrey/ip_autogen/rstmgr/index.html

nbdd0121:
I agree most resets should be in one place, but PoR is special because you need FPGA-specific construct to make it work.

elliotb-lowrisc:
Hmm, I think I should have first clarified what change it is that you're proposing. I had thought you meant the new reset synchronisers should move into `sonata_system`. Do you instead mean that they need to be in a separate reset module at the top-level in order to keep `rst_ctrl` (or whatever has the FPGA-only `initial` statements) to a minimum?

nbdd0121:
Yes.

PS. I am struggling to understand why the separate reset synchronizations are necessary. If we synchronize it to HR clk then the signal should be usable to all resets? They're all derived from a single PLL and therefore should be all synchronized already.

elliotb-lowrisc:
Ah, that makes sense to me now. I'll separate the synchronisation out into a new `rst_sync` module and put that in `top_sonata` in-between `rst_ctrl` and the reset consumers.

You're quite right that these clocks are all synchronous. If only that made getting paths between them to pass timing easy. At the moment there's over 1k failing reset paths to the clk_sys and clk_usb domains. Most of these are to the clk_usb domain, presumably due to the current ratio of clock periods. The easy option I was initially considering was to just use `set_false_path` on these to disable timing checks. No more timing failures! But, I eventually came around to the idea of adding synchronisers instead. These also make the timing failures go away but without disabling checks. Perhaps more importantly, they also prevent metastability and so ensure that everything within a clock domain will come out of reset at the same time. Adrian has observed that Ibex tried to fetch instructions immediately after reset, so for the clk_sys domain it seems important that everything exit reset at once to avoid some logic still being in reset when other logic is trying to use it. For the peripheral domains, it's more just a nice-to-have that also seems to prevent capability exceptions when hammering on the reset button when USB or HyperRAM tests are loaded onto the FPGA. Does that clear things up or make it less understandable?

nbdd0121:
I understand what you tried to achieve but I think the approach should be to hold CPU in reset for more cycles. Usually the way to solve these kind of issue is to be a reset sequence instead of resetting everything at once, e.g. reset peripherals, the interconnects, and finally the core.

Use `prim_flop_2sync` to achieve this without a sequencer is a bit dangerous IMO because it means that the reset is actually *not* synchronous. They'll reach the logic running with fastest clock first, and then slowest clock last.

elliotb-lowrisc:
To clarify, you are talking about the reset _deassertion_ reaching some clock domains ahead of others?

elliotb-lowrisc:
I think synchronising reset deassertion within each clock domain is more important than 1-2 cycle differences between clock domains. The core shouldn’t attempt to access anything other than boot-up instruction memory in its first cycles after reset. This memory should be in the same clock domain, so deasserting the reset across that clock domain cleanly in one cycle seems crucial. Peripherals in other clock domains conversely are accessed some cycles later. They are also accessed via valid-ready asynchronous FIFOs that should ensure that no requests are lost due to the recipient being in reset (so long as they exit reset cleanly).

Not synchronising reset deassertion could lead to multi-cycle differences both within and between clock domains depending on the varying path delays and particular clock edge timings.

GregAC:
> I still think we should have a PoR reset being the only thing in rtl/fpga/rst_ctrl.sv, and have the rest of resets being generated from that one and then we can provide SW interface to the reset the top.

I'd disagree, rst_ctrl, as the name suggest, should control the resets and hence output all of the resets used.

With @elliotb-lowrisc's setup here I think that means we want the synchronizer inside rst_ctrl.

> The core shouldn’t attempt to access anything other than boot-up instruction memory in its first cycles after reset

I am a little nervous about relying on that assumption, however do we actually need it? As you say things are either in the same clock domain and come out of reset at the time or across a clock crossing FIFO. Key question here is how do our CDC FIFOs act when one side is held in reset and the other side has an item written to it? You would hope this is handled properly but we should double check, OT for instance may have made some reset architecture choices that mean this property was never required.

GregAC:
Oh and firmly agreed with @elliotb-lowrisc on the necessity of the separate reset synchronizers.

For CDC and RDC you should do the simplest thing that's easiest to analyze/relies on the fewest assumptions (i.e. avoid assuming the clocks from the PLL are perfectly in sync in this case) and then iterate from there as requirements dictate. Here the overhead of the multiple synchronizers is very modest so no point in optimizing down to a single one even if it looks like it should work fine.

nbdd0121:
> I'd disagree, rst_ctrl, as the name suggest, should control the resets and hence output all of the resets used.

PoR should be separate. On [OpenTitan](https://opentitan.org/book/hw/top_earlgrey/ip_autogen/rstmgr/doc/theory_of_operation.html) the PoR is feed from analog sensor top and then stretched by reset manager and then feed into other resets.

We should do the same here and have the fpga-specific logic just generate that signal, and have the reset handled by rst_ctrl. The rst_ctrl can then be moved into sonata_system.sv, which will make it easier to eventually create SW interface for it.

By having PoR reset and other resets together in the same place will mean that we need to run tilelink signals or a register interface over sonata_system module.

elliotb-lowrisc:
I've had a look and the async primitives:
- prim_fifo_async
  - reset on read-side only: fifo_rptr_gray_q held at zero -> fifo_rptr_sync_q will stay zeroed -> write-side will accept data (full_wclk low -> wready_o high) until full.
  - reset on write-side only: fifo_wptr_gray_q held at zero -> fifo_wptr_sync_combi will stay zeroed -> empty_rclk will stay high -> rvalid_o will stay low.
- prim_sync_reqack in prim_fifo_async_simple with EnRzHs == 1 (used in pulp_riscv_dbg)
  - reset on read-side only: (prim_sync_reqack) dst_fsm_q held at LoSt and dst_req held low -> dst_ack held low -> src_ack low -> src_ack_o low -prim_fifo_async_simple-> pending_d/pending_q can go high if wvalid_i does but cannot go low again -> wready_o will go low cycle after wvalid_i first goes high -> wr_en will only go high once for one cycle -> data from first transaction will be held in data_q indefinitely.
  - reset on write-side only: (prim_sync_reqack) src_fsm_q held at LoSt -> src_req held low -> dst_req held low -> dst_req_o stays low -prim_fifo_async_simple-> dst_req stays low -> rvalid_o stays low.
- prim_fifo_async_simple with EnRzHs == 0 is thankfully not used, as a comment indicates that this would make it not partial-reset-safe.
- prim_fifo_async_sram_adapter is unused.
- prim_flop_2sync provides no flow-control and is therefore a risk.

elliotb-lowrisc:
As to where the synchronisers should live, I think the ease with which we can relocated the synchronisers in future and the relative urgency of fixing the timing failures points to leaving them where they are in the PR for now and moving them if required in future.

elliotb-lowrisc:
Rebase

rtl/ip/hyperram/rtl/hbmc_tl_top.sv
```diff
@@ -155,8 +140,8 @@ module hbmc_tl_top import tlul_pkg::*; #(
       )
       hbmc_arst_sync_idelayctrl
       (
-          .clk   ( clk_i     ),
-          .arst  ( sync_rst   ),
+          .clk   ( ~rst_ni        ),
```
marnovandermaas, line 143: Did you mean for this clock to be connected to reset?

marnovandermaas (commented):
Thanks for putting this together. Have you tested the hyperram after making this change.

rtl/ip/hyperram/rtl/hbmc_tl_top.sv
elliotb-lowrisc: Whoops, no. Fixed now

elliotb-lowrisc: The design worked despite this mistake due to us not using the parameterisation that instantiates it

elliotb-lowrisc:
I can confirm that `hyperram_test` still passes for me with current changes

GregAC:
> PoR should be separate. On [OpenTitan](https://opentitan.org/book/hw/top_earlgrey/ip_autogen/rstmgr/doc/theory_of_operation.html) the PoR is feed from analog sensor top and then stretched by reset manager and then feed into other resets.

Happy for the PoR to be generated outside of rst_ctrl and fed in. Let's open something in JIRA to track refactoring and take this PR as is.

GregAC (approved):

PR #223, opened 2024-10-08
Faster data memory access

alees24:
This ~~_draft_~~ PR collects waivers necessary to achieve an 'UNOPTFLAT warning'-free Verilator build, allowing the removal of two of the `tlul_fifo_sync` modules in sonata_system, and adjustment of the `xbar_main` fabric to achieve single-cycle access to the SRAM/HyperRAM ..._socket_m1 interface.

The build works with Verilator and has been shown to be very nearly timing clean with Vivado (a single violation of -0.087ns occurred) _without_ the changes in commit 820c8cbfbda685a57d58eb012cbf75b5d0c356b3. That commit was excluded from local FPGA builds just to avoid confusion as to the cause of any violations.

The `xbar_main` fabric is modified to introduce latency to the slower peripherals in an effort to reduce pressure on the `cored` host port and aid meeting timing. The IP blocks likely to require lower latency, or benefit from higher performance are left untouched. The main focus has been on reducing latency to data SRAM and HyperRAM. `usbdev` is left untouched because it is a CDC connection and is therefore necessarily already routed through flops.

The additional delays in `xbar_main` are considered to be inconsequential, but they may be over-conservative; a lot of experimentation has been done to reach this point, and some of it may possibly be relaxed.

~~Tidying required:~~

 - ~~the UNOPTFLAT waivers are just added directly to the local RTL files presently; this needs to be achieved more cleanly now that the set of nets has been identified.~~
 - ~~integrity generation and checking on the TL-UL cmd and rsp channels is stripped out using `ifdef presently rather than patched vendoring; this was done because (i) Sonata FPGA does not require integrity, and (ii) more importantly it was responsible for some of the UNOPTFLAT reports. Eliminating it should also help reduce resource pressure.~~

~~Depends on [PR #4](https://github.com/lowRISC/cheriot-ibex/pull/4)~~ Additional commit now includes this.


alees24:
This update is a lot cleaner, migrating UNOPTFLAT waivers into the `sonata_verilator_lint.vlt` file. I think this is preferable to patching the CHERIoT RTL since the warnings about circular logic are an interaction between the core (especially the various bits of logic that drive traffic onto the LSU/data port) and the TL-UL/xbar fabric in Sonata.

If we start introducing waivers into the core RTL itself we run the risk of masking actual combinational loops when the core is deployed elsewhere.

@marnovandermaas There are still a couple of patches to files vendored-in from the cheriot-ibex repo so this remains a draft and I'll address those shortly, but I welcome feedback in the meantime, not least because I still feel rather new to the vendoring/patching flow.

alees24:
This should now be clean for the four permutations of `--trace` present/absent and `+define+RVFI` present/absent, but I fear we may not have heard the last of UNOPTFLAT when the design changes.

marnovandermaas (approved):
This looks good to me. Thanks you for making the bus so much faster to access!

GregAC:
Thanks for this @alees24. I'm wondering if we should upstream the tilelink changes to OpenTitan to enable you to disable integrity given other users of the IP may also want to ditch it?

PR #225, opened 2024-10-09
CHERIoT RTOS version update.

HU90m:
Updated to CHERIoT RTOS to include https://github.com/lowRISC/cheriot-rtos/pull/4

This is the `sonata-system` branch. FYI: All old branches are still available with a number suffix, i.e. `sonata-system-2`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -162,9 +162,9 @@
       cheriotRtosSource = pkgs.fetchFromGitHub {
         owner = "lowRISC";
         repo = "CHERIoT-RTOS";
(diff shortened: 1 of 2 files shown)


marnovandermaas (commented):
Is this a matter of checking that all CHERI checks and tests still work properly?

alees24:
> Is this a matter of checking that all CHERI checks and tests still work properly?

This commit was required for the Sonata SPI CS change (thanks @HU90m) so it has already been merged.

PR #226, opened 2024-10-09
Pinmux no longer buffers

HU90m:
Removed the `inout`s from the `pinmux` and hooked it up in verilator!

In order to do this in a nice way, I rewrote topgen a little (more) and fixed the data format so some assumptions didn't need to be made. I've tried to keep changes to `pinmux.sv` minimal.

*I had used structs and unions but support was patchy, so big unpacked arrays it is.*


HU90m:
Currently passing all merged tests bar the I²C test

elliotb-lowrisc:
Synth 8-6104 warning indeed seem to be gone. :+1:

doc/ip/pinmux.md
```diff
@@ -12,197 +12,197 @@ The default value for all of these selectors is `'b10`.
...
+| 0x833 | `gpio[2].ios[13]` | 0, `pmod1[5]` |
+| 0x834 | `gpio[0].ios[14]` | 0, `rph_txd0` |
+| 0x835 | `gpio[1].ios[14]` | 0, `ah_tmpio14` |
+| 0x836 | `gpio[2].ios[14]` | 0, `pmod1[6]` |
+| 0x837 | `gpio[0].ios[15]` | 0, `rph_rxd0` |
+| 0x838 | `gpio[1].ios[15]` | 0, `ah_tmpio15` |
+| 0x839 | `gpio[2].ios[15]` | 0, `pmod1[7]` |
+| 0x83a | `gpio[0].ios[16]` | 0, `rph_g16_ce2` |
+| 0x83b | `gpio[1].ios[16]` | 0, `ah_tmpio16` |
+| 0x83c | `gpio[2].ios[16]` | 0 |
```
marnovandermaas, line 160: Hmmm, what about the default values here? I kind of wanted the software to be able to write 2 to all selectors to make things default.

rtl/system/sonata_system.sv
```diff
@@ -1043,16 +1015,19 @@ module sonata_system
...
 
-  logic [WordWidth-1:0] gpio_array_inputs [GPIO_NUM + 1];
-  logic [WordWidth-1:0] gpio_array_outputs [GPIO_NUM + 1];
-  logic [WordWidth-1:0] gpio_array_output_enables [GPIO_NUM + 1];
+  // 0: General Purpose
+  // 1: Raspberry Pi HAT g0-15
+  // 2: Raspberry Pi HAT g16-27
+  // 3: Arduino Shield 0-15
+  // 4: Arduino Shield 16-17
+  // 5: PMOD
```
marnovandermaas, line 1023: This is now no longer true:
0: general purpose
1: Raspberry Pi
2: Arduino Shield
3: Pmod

marnovandermaas (commented):
Some initial comments, but I wasn't able to go through the whole of the last commit. It's a big change.

util/top_gen/parser.py
```diff
@@ -75,6 +86,19 @@ def check_pins(pins: list[Pin]) -> list[Pin]:
             raise ValueError("All pins must have unique names.")
         return pins
 
+    @field_validator("blocks")
+    @staticmethod
+    def check_blocks(blocks: list[Block]) -> list[Block]:
+        block_names = {block.name for block in blocks}
+        expected_block_names = {"gpio", "i2c", "spi", "uart"}
```
marnovandermaas, line 107: Is there a reason why you are hard-coding this? If we added PWM for example we would have to add it in the toml and also here.

dv/verilator/top_verilator.sv
```diff
@@ -156,114 +153,32 @@ module top_verilator (input logic clk_i, rst_ni);
...
+  sonata_in_pins_t in_from_pins;
+  sonata_out_pins_t out_to_pins;
+  sonata_inout_pins_t inout_from_pins, inout_to_pins, inout_to_pins_en;
 
-  // Stub of pinmux to make sure it keeps compiling.
-  pinmux u_pinmux (
-    .clk_i,
-    .rst_ni,
+  assign uart_sys_tx = out_to_pins[OUT_PIN_SER0_TX];
+  assign uart_aux_tx = out_to_pins[OUT_PIN_SER1_TX];
```
marnovandermaas, line 163: What does `aux` mean here?

dv/verilator/top_verilator.sv
```diff
@@ -398,8 +294,12 @@ module top_verilator (input logic clk_i, rst_ni);
     .hyperram_nrst(),
     .hyperram_cs  (),
 
-    .tl_pinmux_o (),
-    .tl_pinmux_i (tlul_pkg::TL_D2H_DEFAULT)
+    .in_from_pins_i     (in_from_pins    ),
+    .out_to_pins_o      (out_to_pins     ),
+    .out_to_pins_en_o   (                ),
```
marnovandermaas, line 307: Although not necessary to make a tri-state signal, it might still be useful to have this visible in simulation.

rtl/fpga/top_sonata.sv
```diff
@@ -230,121 +232,16 @@ module top_sonata (
   assign usrusb_oe  = !dp_en_d2p;  // Active low Output Enable.
   assign usrusb_sus = !rx_enable_d2p;
 
+  sonata_in_pins_t in_from_pins;
+  sonata_out_pins_t out_to_pins, out_to_pins_en;
+  sonata_inout_pins_t inout_from_pins, inout_to_pins, inout_to_pins_en;
+
+  wire sonata_out_pins_t   output_pins;
+  wire sonata_inout_pins_t inout_pins;
```
marnovandermaas, line 240: We tend to use `logic` instead of `wire`. Assigning combinatorially to a `logic` essentially makes it a wire.

rtl/fpga/top_sonata.sv
```diff
@@ -425,6 +322,7 @@ module top_sonata (
                       usrLed, // User LEDs (8 bits)
                       lcd_backlight, lcd_dc, lcd_rst, gp_lcd_cs // LCD screen
                     }),
+    .gp_o_en        (),
```
marnovandermaas, line 325: Thanks :)

util/top_gen/generator.py
```diff
@@ -3,322 +3,359 @@
...
+                                instance,
                                 io.name,
-                                inst_idx,
-                                bit_str,
-                                io.type == Direction.INOUT,
-                            )
+                                io_index,
+                            ),
+                            direction=io.type,
+                            default_value=io.default,
```
marnovandermaas, line 174: Nit: default value above direction.

util/top_gen/generator.py
```diff
@@ -3,322 +3,359 @@
...
-    block_connections = block_connections_map(config)
-    pin_to_block_outputs = pin_to_block_output_map(config, block_connections)
+        return (block_type_idx, io_index, uid.instance)
 
-    input_pins = list(input_pins_iter(config, block_connections))
-    output_pins = list(output_pins_iter(config, pin_to_block_outputs))
-    inout_pins = list(
-        inout_pins_iter(config, block_connections, pin_to_block_outputs)
-    )
+    output_block_ios.sort(key=legacy_ordering)
```
marnovandermaas, line 343: Is this so that we can change the ordering later?

marnovandermaas (commented):
I think this PR does what it says. I've left a few comments as I was going through.

This is not something you need to do in this PR, but in the future it would have been nice to have the last commit split up into multiple commits. For example: extracting the buffering from the pinmux, rewriting the parsing and changing the named mapping to the packed array. This is mainly to help reviewing.

dv/verilator/top_verilator.sv
HU90m: Auxiliary I think, this pre-dates the PR and I was keep changes minimal. I would like to remove this as well.

util/top_gen/parser.py
HU90m: And also to the templates and RTL. I think this is a good sanity check to have.

util/top_gen/generator.py
HU90m: Yep

rtl/fpga/top_sonata.sv
GregAC: Three things here:

1. `sonata_out_pins_t` and `sonata_inout_pins_t` are already `logic` from the `typedef`, maybe not hugely surprising Verilog apparently allows you to tag a `wire` on as well (or at least Vivado/Verilator aren't complaining about it) but not something we should actually utilize. So we should not use `wire` with these types as is
2. I'd have a preference for using `wire` in the particular case of nets that can be tri-stated. `logic` can be state holding element which we definitely can't have for tri-state so `wire` is a better statement of intent as well as pointing out there's something a bit unusual about this net (given we basically use `logic` everywhere). This is also clearly stated in our style guide: https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#use-logic-for-synthesis
> There are exceptions for places where logic is inappropriate. For example, nets that connect to bidirectional (inout) ports must be declared with wire. 
3. It may be better to never declare nets like this that can be tri-state at all, instead preferring to confine tri-state nets entirely to module ports that have `'inout` so it's always clear these are tri-state nets. This can be achieved by having a `padring` module that contains the pad interface logic ala opentitan

https://github.com/lowRISC/opentitan/blob/3e3107c7c651612b73d7d7a0004aee7fc5737664/hw/top_earlgrey/rtl/autogen/chip_earlgrey_asic.sv#L307-L753

Note the concatenation syntax that allows you to have a bunch of named inout pins get put into an array you can use within the module.

rtl/fpga/top_sonata.sv
```diff
@@ -553,4 +429,100 @@ module top_sonata (
...
+  end
+  for (genvar idx = 0; idx < INOUT_PIN_NUM; ++idx) begin
+    // Would use IOBUF, Vivado seems to want me to add (* io_buffer_type = "none" *)
+    // to each of the inout leaving the module if I do.
+    assign inout_pins[idx]      = inout_to_pins_en[idx] ? inout_to_pins[idx] : 1'bz;
+    assign inout_from_pins[idx] = inout_pins[idx];
+  end
+
+  // Breaking out pins
+  assign output_pins[OUT_PIN_SER0_TX]       = ser0_tx;
```
GregAC, line 445: Isn't this the wrong way round? `ser0_tx` is an output pin and `output_pin[OUT_PIN_SER0_TX]` is the value it should be driven to

rtl/fpga/top_sonata.sv
```diff
@@ -553,4 +429,100 @@ module top_sonata (
     .rst_usb_no (rst_usb_n),
     .rst_hr_no  (rst_hr_n)
   );
-endmodule
+
+  // Pin Buffers
+  for (genvar idx = 0; idx < OUT_PIN_NUM; ++idx) begin
+    assign output_pins[idx] = out_to_pins_en[idx] ? out_to_pins[idx] : 1'bz;
```
GregAC, line 435: If this is intended to drive top-level `output` pins as the name suggests (i.e. specifically not `inout` pins) then it should be not tri-stated. Instead it can just be 0 when output isn't enabled.

There is an argument that we never have any top-level `output` pins and they're all `inout` of course but I think for now given we do have top-level `output` we just drive to 0 when they're not enabled.

rtl/fpga/top_sonata.sv
```diff
@@ -553,4 +429,100 @@ module top_sonata (
...
+  end
+  for (genvar idx = 0; idx < INOUT_PIN_NUM; ++idx) begin
+    // Would use IOBUF, Vivado seems to want me to add (* io_buffer_type = "none" *)
```
GregAC, line 438: Ultimately this is an argument for have specific pad instances that have the top-level IO directly connected through to (as in using exclusively port IO no intermediate `wire` nets that aren't part of a port declaration). Within that pad instance you can do then do whatever tool/technology specific things you need to do. That's what's done in OpenTitan.

GregAC (commented):
Thanks for this @HU90m I think this is almost there just the stuff around `logic` v `wire` and exactly how you deal with the tri-stating nets as I've commented on that needs sorting out.

Ultimately I'd like to use a pad wrapper module exactly as we do in OpenTitan but it's not needed for this PR.

Please make sure this gets some testing on FPGA using @AlexJones0's pinmux testing work before it gets merged.

rtl/fpga/top_sonata.sv
GregAC: Of course you could define in the pinmux config the default state of an output when it's not connected to anything but I think best to minimise new stuff in this PR.

rtl/fpga/top_sonata.sv
HU90m: [I did have a padring module that used the OT pad wrapper](https://github.com/HU90m/sonata-system/commit/414c6d33319a784dd84779f6ca3bb064d492dc84#diff-36f2138a2eccdc3286e62dc9c05fd946567aa9a18831a5e9d3b864339e7fd8e8), but came across this issues.

I agree this would be very nice to have this but as you'd guessed I was trying to keep the changes in this PR as minimal as possible. Definitely do want something like this in the future.

rtl/fpga/top_sonata.sv
GregAC: Ah good point, I keep forgetting these details!

In which case `wire` in front of these type should be fine and in line with the style guide but vital we use `wire` (we don't want an implicit `var`).

Though you may run into issues using this nonetheless. Ultimately tri-state drivers and nets do not exist in the FPGA, only at the very edge at the pad instance. You need the synthesis tool to realise you're creating a piece of tri-state logic that can be directly implemented by the tri-state driver available to it (the one in a pad) that's then directly connected out to the pin. It should be able to see this through a wire connect from an IO to another module but a possibility it doesn't (may explain some of the difficulties you had with vivado and using the OT pad wrappers).

dv/verilator/top_verilator.sv
marnovandermaas: Fair enough

dv/verilator/top_verilator.sv
marnovandermaas: Ah, didn't know this predates the PR. Resolving this as it is not essential.

util/top_gen/parser.py
marnovandermaas: Ok, I assume you put this in because you ran into issues, so let's resolve this for now.

data/pins_sonata.xdc
```diff
@@ -84,27 +84,27 @@ set_property -dict { PACKAGE_PIN K2  IOSTANDARD LVCMOS18 } [get_ports {usrusb_su
...
-set_property -dict { PACKAGE_PIN D17 IOSTANDARD LVCMOS33 } [get_ports {pmod0[6]}]
-set_property -dict { PACKAGE_PIN K13 IOSTANDARD LVCMOS33 } [get_ports {pmod0[7]}]
+set_property -dict { PACKAGE_PIN H14 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[0]}]
+set_property -dict { PACKAGE_PIN F16 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[1]}]
+set_property -dict { PACKAGE_PIN F15 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[2]}]
+set_property -dict { PACKAGE_PIN G14 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[3]}]
+set_property -dict { PACKAGE_PIN J13 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[4]}]
+set_property -dict { PACKAGE_PIN E17 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[5]}]
+set_property -dict { PACKAGE_PIN D17 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[6]}]
+set_property -dict { PACKAGE_PIN K13 IOSTANDARD LVCMOS33 IO_BUFFER_TYPE NONE } [get_ports {pmod0[7]}]
```
GregAC, line 94: Is the `IO_BUFFER_TYPE NONE` required because the pad primitive gives us a an explicit buffer instantiation?

rtl/fpga/top_sonata.sv
GregAC: Still need to assign to `0` when disabled here.

rtl/fpga/padring.sv
```diff
@@ -0,0 +1,34 @@
...
+  output logic [InoutNumber-1:0] inout_from_pins_o,
+
+  inout  logic [InoutNumber-1:0] inout_pins_io
+);
+  prim_pad_wrapper_pkg::pad_attr_t pad_attr = '{
+    default:'0,
+    drive_strength: '1
+  };
+
+  prim_pad_wrapper u_inout_pad[InoutNumber] (
```
GregAC, line 19: First time I've seen this syntax (parameter in square brackets after a module name) but if it works it works :)

GregAC (approved):
Ok this is looking good from my perspective, thanks @HU90m 

Approving now to unblock but a few things before you merge

- Approval from @marnovandermaas 
- Output default to `0` when not enabled rather than `z`
- Ensure this is tested on a real board use @AlexJones0's pinmux tests.

HU90m:
> * Output default to `0` when not enabled rather than `z`

Ah thank you! I forgot about this

rtl/fpga/padring.sv
HU90m: Yeah, I really like module arrays module arrays. I think they are pretty neat and a little more semantic than generator loops.

data/pins_sonata.xdc
HU90m: Exactamundo, you can put `(* io_buffer_type = "none" *)` attributes on the top level ports in the SystemVerilog, but I thought it was neater having them in the TCL.

AlexJones0 (commented):
Can confirm from manual testing that after rebasing #227 and #230 on this branch that both the automated pinmux tests and the manual pinmux muxing check are still passing on FPGA.

HU90m:
> Can confirm from manual testing that after rebasing #227 and #230 on this branch that both the automated pinmux tests and the manual pinmux muxing check are still passing on FPGA.

Yay! thanks Alex

marnovandermaas (approved):
Let's get this in and iterate over it with follow up PRs.

PR #227, opened 2024-10-10
Add CHERI Pinmux tests

AlexJones0:
This PR is ready, but it is marked as a *draft* for a few reasons:
 1. It depends on #212 being merged. This PR uses and changes the driver made there. ***For now, this PR contains the 3 commits from that PR - these should be removed before merging.***
 2. It will fail in simulation - mocks need to be set up connecting the GPIO via the 3 connections used in these Pinmux tests.
 3. Pinmux does not currently work in Verilator - and thus CI test results are meaningless regardless of (2).

This PR does the following:
  - Adds more UART block definitions to [`defs.h`](https://github.com/lowRISC/sonata-system/blob/main/sw/common/defs.h) and [`sonata-devices.hh`](https://github.com/lowRISC/sonata-system/blob/main/sw/cheri/common/sonata-devices.hh).
  - Adds Pinmux options for UART3 RX/TX over Arduino Shield D0 and D1, to allow Pinmux testing to take place.
  - Updates the pinmux driver (from #212) to reflect the above.
  - Adds a pinmux test suite containing 5 tests: GPIO, UART, I2C, SPI and Mux tests.

See the commit messages for more detailed information.

The 5 tests can be summarised as follows:
  - **GPIO**: requires a manual connection between Ardunio Shield D8 and D9. Writes data to one pin and reads from another, when muxed to GPIO. Also checks that this fails when not muxed to GPIO.
  - **UART**: requires a manual connection between Mikrobus (P7) RX and TX. Sends data over one pin and reads from the other, when muxed to UART. Also checks that this fails when not muxed to UART.
  - **SPI**: tests that SPI works by reading the JEDEC ID of the Application SPI Flash on the board. Also checks that this fails when not muxed.
  - **I2C**: requires a Raspberry Pi Sense HAT to be attached to the board. Tests that I2C works by running the existing I2C RPi Sense HAT `ID_EEPROM` and `WHO_AM_I` tests. Also checks that this fails when not muxed.
  - **Mux**: requires a manual connection between Arduino Shield D0 and D1. Tests the actual muxing functionality, rather than just enabling/disabling via pinmux. First muxes the pins as UART, and checks the UART test passes and the GPIO test fails. Then muxes as GPIO, and checks the opposite is true. Then muxes back and checks once more. This will hence give us a test to run in CI showing us muxing between pins using different blocks.

The number of test iterations can be configured via compilation flags. This number is purposefully kept at `1` to keep runtime reasonable on Verilator.

This can be run via the standard test running procedure (building and using `mem_helper.sh` on the `test_runner` target or using the helper script) to confirm that it works. Note the three connections on the board (and the Raspberry Pi Sense HAT) required for the tests to pass. These connections are used so that we are able to test GPIO/UART (and thus the muxing functionality) using just the board, without any sort of secondary device and the complexity that would entail.

This has been tested and has passed consistently for 100 iterations on FPGA.


GregAC (commented):
LGTM thanks @AlexJones0 though should get a SW person to review and approve

AlexJones0:
Push was just a rebase on main, and dropped the commit that adds additional UART blocks as that was already merged in a different PR (see #219).

AlexJones0:
Force push rebases on main (including removing the now-merged commits from #212)

AlexJones0:
Push rebased the PR on top of main, which now also includes the logging changes (which have been updated for this test), includes the I2C driver fixes required to hopefully make this test pass, and includes the Verilator pin connections required to hopefully make this pass in simulation.

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -0,0 +1,546 @@
...
+
+/**
+ * Attempt to retrieve the JEDEC ID of the Spi Flash device on the Sonata board,
+ * and compare it to the known (expected) JEDEC ID. This is separate from the
+ * similar test in the SPI tests, as it skips the initialisation logic that
+ * should not be repeated in the pinmux check.
+ *
+ * Returns the number of failures during the test.
+ */
+static int spi_jedec_id_test(Capability<volatile SonataSpi> spi, SpiFlash spi_flash) {
```
HU90m, line 204: `spi` doesn't appear to be used in this function.

HU90m (approved):
I'll merge now, so you don't have to worry about rebasing again, but I've left some comments that would be nice to address.

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -0,0 +1,226 @@
...
+  int failures = 0;
+
+  // Read the JEDEC ID from Flash
+  uint8_t jedec_id[3] = {0};
+  spi_flash.read_jedec_id(jedec_id);
+
+  // Check that the retrieved ID matches our expected value
+  for (size_t index = 0; index < 3; index++) {
```
engdoreis, line 212: `3 -> sizeof(jedec_id)`

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -0,0 +1,546 @@
...
+  SonataPinmux Pinmux                 = SonataPinmux(pinmux);
+
+  // Create bounded capabilities for other devices, to be used in testing.
+  Capability<volatile SonataSpi> spi = root.cast<volatile SonataSpi>();
+  spi.address()                      = SPI_ADDRESS;
+  spi.bounds()                       = SPI_BOUNDS;
+
+  Capability<volatile SonataGpioFull> gpio_full = root.cast<volatile SonataGpioFull>();
+  gpio_full.address()                           = GPIO_ADDRESS;
+  gpio_full.bounds()                            = GPIO_FULL_BOUNDS;
```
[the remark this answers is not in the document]
AlexJones0: For `SonataSpi` I agree, for `SonataGpioFull` this is essentially a small temporary version of the driver sitting in a PR at https://github.com/lowRISC/cheriot-rtos/pull/13, so I think when that is merged and we bump the revision of CHERIoT RTOS used in `sonata-system` we should update all CHERI GPIO usages (including this) to go through that.

sw/cheri/tests/pinmux_tests.hh
AlexJones0: Looks like I somehow lost the logic that (re)configures the SPI along the way, will add that now.

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -0,0 +1,226 @@
...
+ * Returns the number of failures during the test.
+ */
+static int pinmux_spi_flash_test(SonataPinmux *pinmux, Capability<volatile SonataSpi> spi, SpiFlash spi_flash) {
+  constexpr uint8_t PmxSpiFlashDataToSpiTx0   = 1;
+  constexpr uint8_t PmxSpiFlashClockToSpiClk0 = 1;
+
+  int failures = 0;
+
+  // Ensure the SPI Flash pins are enabled using Pinmux
+  if (!pinmux->output_pin_select(SonataPinmux::OutputPin::SpiFlashData, PmxSpiFlashDataToSpiTx0)) failures++;
```
engdoreis, line 279: This is my personal preference, I don't like if statements without `{}`, so I would suggest:
failures += (condition == true);

sw/cheri/tests/pinmux_tests.hh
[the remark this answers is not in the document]
AlexJones0: From checking the code now it appears that I kept `SonataSpi` here because the `SpiFlash` util constructor expects a capability `Capability<volatile SonataSpi>` to the SPI, whereas the `spi_ptr` function only returns a pointer. I'm not sure if there is an easy way to make that conversion - please let me know if there is and I'll happily change this.

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -0,0 +1,546 @@
...
+/**
+ * Sets the output enable value of some GPIO pin. Setting this to 1 places the
+ * pin in output mode, and to 0 places it in input mode.
+ */
+static inline void set_gpio_output_enable(Capability<volatile SonataGpioFull> gpio, GpioPin pin, bool value) {
+  const uint32_t bit_mask    = (1 << pin.bit);
+  const uint32_t write_value = ((value ? 1 : 0) << pin.bit);
+  uint32_t output_enable     = gpio->instances[pin.instance].output_enable;
+  output_enable &= ~bit_mask;
+  output_enable |= write_value;
```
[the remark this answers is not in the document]
AlexJones0: This is not equivalent logic, is it? `~bit_mask` will always be all 1s except a 0 in the bit being written to, so `write_value` (which is all 0s in bits not being written to, and either a 0 or 1 in the bit being written to) when `&`ed together with `~bit_mask` will always output a `0` for that bit.

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -29,19 +29,80 @@ using namespace CHERI;
...
+  gpio->instances[pin.instance].output = output;
+}
+
+/**
+ * Sets the output enable value of some GPIO pin. Setting this to 1 places the
+ * pin in output mode, and to 0 places it in input mode.
+ */
+static inline void set_gpio_output_enable(Capability<volatile SonataGpioFull> gpio, GpioPin pin, bool value) {
+  const uint32_t bit_mask    = (1 << pin.bit);
+  const uint32_t write_value = ((value ? 1 : 0) << pin.bit);
```
engdoreis, line 89: It's safe to cast bool to integer in C++.
```suggestion
const uint32_t write_value = value  << pin.bit;
```

engdoreis (commented):
LGTM.

AlexJones0:
See #278 which addresses remaining comments.

sw/cheri/tests/pinmux_tests.hh
AlexJones0: This makes sense, I'll convert these kinds of statements in a follow-up PR

PR #229, opened 2024-10-10
[sw, utils] Add a lightweight c++23 print like library for logging

engdoreis:
Add [reisfmt](https://github.com/engdoreis/reisfmt) a lightweight c++23 like print implementation that follows the specification in https://en.cppreference.com/w/cpp/utility/format/spec


sw/cheri/common/console.hh
```diff
@@ -0,0 +1,42 @@
...
+#include "sonata-devices.hh"
+#include "../../common/defs.h"
+#include <platform-uart.hh>
+
+#include "console.hh"
+
+#define CC_BOLD "1"
+#define CC_RED "31"
+#define CC_GREEN "32"
+#define CC_RESET "0"
```
nbdd0121, line 18: I think I'll probably still want them to have the full escape code and them just write them directly.

engdoreis: I agree, but I want this PR to only translate the old way of logging to print. I'm happy to tackle that in a follow-up PR.

nbdd0121 (approved):

PR #230, opened 2024-10-10
Add manual pinmux testing software and configure PMOD pinmuxing

AlexJones0:
This PR does the following:
  - Separates out the PMOD pins in the `top_config.toml` Pinmux logic so that each individual pin can mux differently.
  - Adds standard mux options for PMOD0 and PMOD1 based on the [Digilent PMOD Interface Specification](https://digilent.com/reference/_media/reference/pmod/pmod-interface-specification-1_2_0.pdf).
  - ~~Pulls up the PMOD pins that can be muxed for I2C, as they may not be externally driven~~ **does not do this any more, see comment by @GregAC below**.
  - ~Adds more SPI block definitions to [`defs.h`](https://github.com/lowRISC/sonata-system/blob/main/sw/common/defs.h)~ **removed as merged by different PR**.
  - Adds a pinmux testplan checker which can be used to define and execute pinmux testplans that run tests of the Pinmux with manual intervention. See the commit message of 00759a0415d7692f468c0bca5a8d8d8fd1a0c489 for a more complete explanation of what this does.
  - Adds an "all blocks" pinmux check which tests dynamically muxing between using GPIO, I2C, SPI and UART on the same PMOD0 pinmuxed pins.

As of recent reviews it now also:
  - Bumps the version of CHERIoT RTOS being used to include the updated GPIO driver - [see the differences between the two revisions here](https://github.com/CHERIoT-Platform/cheriot-rtos/compare/bcd25c8734edb5395ac239b16af336987da45e9d...8fbbda7b0ee528764892e1add9b539f9ec9064b0).
  - Updates the GPIO definitions (in [`defs.h`](https://github.com/lowRISC/sonata-system/blob/main/sw/common/defs.h))  to match the current RTL/driver.
  - Refactors out the common GPIO functionality and UART/GPIO functionality that is shared between this PR and the pinmux tests into its own `block_tests` shared library that is used by both, to reduce code duplication.

See the commit messages for more detailed information.
**It's probably best to start looking [here](https://github.com/lowRISC/sonata-system/blob/c78a04ef57ee0d816a0a4f1fe3764e2173c74100/sw/cheri/checks/pinmux_all_blocks_check.cc#L98) to get an idea of exactly what this PR tries to achieve.**

The interface for this is currently not quite as nice as I would like: you have to define nested structs separately and then assign them rather than using nested initialisers, and a lot of arguments are being passed to some functions. I also don't seem to be able to use a union to save space. This is primarily because of issues caused by the lack of `memcpy` - the interface for using the checker would be much nicer if we could copy structs but without that I've tried to make the interface as nice as possible.

To run this test (and any other user-defined testplans that test SPI/I2C), a PMOD SF3 flash and PMOD Colour sensor are required on hand, so that they can be plugged in to be used in tests. UART and GPIO tests simply require a wire to connect the two pins used in a loopback. There is still a test to check the now unused BH1745 luminance and colour sensor, but this isn't being used for the "all blocks" check. This can be run via the standard procedure (building and using `mem_helper.sh` on the `pinmux_all_blocks` target) to confirm that it works.

This has been tested to be working manually on an FPGA, and the RTL changes have been tested to pass on FPGA running the existing tests with the pinmux test defined in #227. **Ideally merge that PR before this PR though, and re-run the CI tests with pinmux to be 100% sure.**


AlexJones0:
Push was just a rebase on main, and dropped the commit from #227 that adds additional UART blocks as well as the commit from this PR that adds additional SPI blocks, as those were already merged in a different PR (see https://github.com/lowRISC/sonata-system/pull/219)

data/pins_sonata.xdc
```diff
@@ -92,6 +92,9 @@ set_property -dict { PACKAGE_PIN J13 IOSTANDARD LVCMOS33 } [get_ports {pmod0[4]}
 set_property -dict { PACKAGE_PIN E17 IOSTANDARD LVCMOS33 } [get_ports {pmod0[5]}]
 set_property -dict { PACKAGE_PIN D17 IOSTANDARD LVCMOS33 } [get_ports {pmod0[6]}]
 set_property -dict { PACKAGE_PIN K13 IOSTANDARD LVCMOS33 } [get_ports {pmod0[7]}]
+## Enable pull-ups because this I2C bus will often be undriven externally.
+set_property PULLTYPE PULLUP [get_ports {pmod0[0]}]
```
GregAC, line 96: Please do not add pull ups to the PMOD pins. They've got multiple uses only one of which is GPIO so a permanent pull up would be a surprising property. I've just had a look at the schematic for the digilent colour detector pmod: https://uk.farnell.com/digilent/410-348/pmod-colour-detector-dev-board/dp/2893222 and that has pinups on the board (which you can disable by removing some jumpers). It's pretty common for I2C breakouts to have their own pull-ups (all the Qwiic boards do for instance) and I think it's the r-pi hat that's the odd one out here where it expects I2C pull-ups on the main board (and the latest version of the Sonata board includes physical pin-ups as the FPGA ones are very weak).

AlexJones0: This makes sense, I have removed the pull ups now.

AlexJones0:
Resolved all remaining issues so now marking ready for review, please see original edited comment on this PR.

sw/cheri/checks/pinmux_checker.hh
```diff
@@ -0,0 +1,163 @@
...
+   * `memcpy` makes that very difficult, i.e.
+   *   union TestData {
+   *     UartTest uart;
+   *     GpioTest gpio;
+   *     I2cTest i2c;
+   *     SpiTest spi;
+   *   } test_data;
+   */
+  // The data to use in the test. The attribute corresponding to the specified `type`
+  // attribute must be filled in with appropriate data.
```
HU90m, line 152:
```suggestion
  /**
   * The data to use in the test. The attribute corresponding to the specified `type`
   * attribute must be filled in with appropriate data.
   * Ideally we would union the below attributes to save space, but the lack of
   * `memcpy` makes that very difficult, i.e.
   *   union TestData {
   *     UartTest uart;
   *     GpioTest gpio;
   *     I2cTest i2c;
   *     SpiTest spi;
   *   } test_data;
   */
```

sw/cheri/checks/pinmux_checker.cc
```diff
@@ -0,0 +1,495 @@
...
+ * the Tx and the Rx of the specified UART to be manually connected on the
+ * board, so that data that is written can be read back by the board.
+ *
+ * @param read_timeout_msec The timeout in milliseconds provided to read each
+ * individual byte sent over UART. Any timeout causes immediate failure.
+ * @param test_length The number of random bytes to try sending over UART.
+ *
+ * Returns true if the test passed, or false if it failed.
+ */
+static bool uart_send_receive_test(ds::xoroshiro::P32R8 &prng, UartPtr uart, uint32_t read_timeout_msec,
```
HU90m, line 62: This is also exists in the pinmux test. Would you mind making a common library for shared functions.

sw/cheri/checks/pinmux_checker.cc
```diff
@@ -0,0 +1,495 @@
...
+  output_enable |= write_value;
+  gpio->instances[instance].output_enable = output_enable;
+}
+
+/**
+ * Get the value of some GPIO input pin. This will only be visible if the pin
+ * has its corresponding `output_enable` bit set to 0, so that the pin is in
+ * input mode.
+ */
+inline bool get_gpio_input(Capability<volatile SonataGpioFull> gpio, GpioPin pin) {
```
HU90m, line 44: I've merged https://github.com/lowRISC/cheriot-rtos/pull/13 , so you can use the new driver functions now.

HU90m (changes requested):

sw/cheri/checks/pinmux_checker.hh
AlexJones0: Side comment: Ideally I would figure out a more object-oriented approach using C++'s classes here but I spent quite a while battling the lack of `memcpy` and so if this is a change we think is important, I can look at iterating on this later down the line.

sw/cheri/checks/pinmux_checker.cc
AlexJones0: Sure! This was just duplicated because I didn't want the PRs to be dependent upon each other, making a longer critical path for reviews. Now that the pinmux test PR is merged I'm happy to migrate this to a common library.

AlexJones0:
Latest push rebases on main and addresses comments.

AlexJones0:
Latest pushes were me trying to fix failures in CI (I just forgot to include the new library files in the commit apparently :sweat:)

sw/cheri/common/block_tests.hh
```diff
@@ -0,0 +1,45 @@
...
+
+#pragma once
+#include "../common/sonata-devices.hh"
+#include <cheri.hh>
+#include <platform-gpio.hh>
+#include <ds/xoroshiro.h>
+
+using namespace CHERI;
+
+// The full definition of Sonata's GPIO aray.
```
HU90m, line 15:
```suggestion
// The full definition of Sonata's GPIO array.
```

sw/cheri/tests/pinmux_tests.hh
```diff
@@ -490,7 +330,7 @@ void pinmux_tests(CapRoot root, Log &log) {
 
   Capability<volatile SonataGpioFull> gpio_full = root.cast<volatile SonataGpioFull>();
   gpio_full.address()                           = GPIO_ADDRESS;
-  gpio_full.bounds()                            = GPIO_FULL_BOUNDS;
+  gpio_full.bounds()                            = GPIO_BOUNDS * GPIO_NUM;
```
HU90m, line 333:
```suggestion
  gpio_full.bounds()                            = GPIO_RANGE * GPIO_NUM;
```

sw/cheri/common/block_tests.hh
```diff
@@ -0,0 +1,45 @@
...
+struct SonataGpioFull {
+  SonataGpioGeneral general;
+  SonataGpioRaspberryPiHat rpi;
+  SonataGpioArduinoShield arduino;
+  SonataGpioPmod pmod;
+};
```
HU90m, line 21: Struct of capabilities would be future proof

HU90m (approved):

sw/cheri/common/block_tests.hh
AlexJones0: Changed to this, along with appropriate use of `GPIO_BOUNDS` and `GPIO_RANGE` now so that this is resilient to gaps in the GPIO registers being introduced in the future.

sw/cheri/tests/pinmux_tests.hh
AlexJones0: See below comment; this change is now redundant due to the change to using a struct of capabilities instead.

HU90m (approved):

PR #231, opened 2024-10-10
Improve robustness of test_runner

alees24:
Ensure that `intr_test` for Status-type interrupts is cleared after using it within `PlicTest`, because otherwise subsequent tests may fail.
Helps `test_runner` run multiple times without an intervening system-global reset, and better supports test reordering.


marnovandermaas (approved):
This looks good to me. Maybe good to give Alex or Hugo a chance to review before merging.

alees24:
I've pushed a second commit, having tracked down and resolved the continuously-asserted interrupt that occurs with repeated downloads/runs of the test_runner program without intervening IP block reset. This will also permit tests to be executed after the `plic_tests` function.

PR #232, opened 2024-10-11
Relax td_i/tms_i hold requirement

elliotb-lowrisc:
Sole remaining timing failure is a setup timing path from the td_i port. Fix it by relaxing hold timing, which appeared to be the root cause of the setup timing issue.
We can justify  this relaxing of the hold timing constraint by assuming that the driver of the input is launching the data on falling clock edges as they are supposed to.


marnovandermaas (approved):
Looks good to me.

PR #235, opened 2024-10-13
SPI loopback, software reset and associated testing

alees24:
This PR consists of three commits:

1. An implementation of a loopback test using an external jumper wire on the RPi header (P19-P20) which could be installed in CI, although this is proof-of-concept and we probably want to use one of the other SPI connections since the RPi header is probably in use.
2. Implement an internal loopback mode in the SPI controller, to permit testing of all controllers in all modes (CPOL, CPHA, MSB_FIRST) and a software reset mechanism if - say - a command is started but not completed, and a block reset has not been/cannot be performed.
3. Test all of the SPI controllers using internal loopback, and all modes.

This has all been tested on FPGA; I suggest that once there's agreement on the physical practicalities w.r.t. CI we should enable a loopback test on a suitable connector. For now the RPi header-based external loopback test is excluded because the I2C testing uses the Sense HAT.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -38,7 +38,7 @@ in {
     doCheck = true;
     buildInputs = [sonata-simulator pythonEnv];
     checkPhase = ''
-      python ${../util/test_runner.py} -t 60 sim \
+      python ${../util/test_runner.py} -t 120 sim \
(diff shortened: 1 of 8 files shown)


GregAC (approved):
Thanks @alees24 useful extra features here and always good to get some more tests.

PR #237, opened 2024-10-14
40 MHz Core clock

elliotb-lowrisc:
Increase `clk_sys` from 30 MHz to 40 MHz. Includes some physical-design-like tweaks to the Vivado flow to achieve timing closure at 40 MHz or possibly beyond.

See individual commit messages for further details.

### DIFF
--- a/dv/verilator/top_verilator.sv
+++ b/dv/verilator/top_verilator.sv
@@ -9,7 +9,7 @@ module top_verilator (input logic clk_i, rst_ni);
   parameter bit DisableHyperram = 1'b0;
 
   // System clock frequency.
-  localparam int unsigned SysClkFreq = 30_000_000;
+  localparam int unsigned SysClkFreq = 40_000_000;
   // HyperRAM clock frequency.
   localparam int unsigned HRClkFreq  = 100_000_000;
   localparam int unsigned BaudRate   = 921_600;
--- a/nix/software.nix
+++ b/nix/software.nix
@@ -11,9 +11,9 @@
   cheriotRtosSource = pkgs.fetchFromGitHub {
     owner = "lowRISC";
     repo = "CHERIoT-RTOS";
-    rev = "c29bdeccc0335b81acea09c069fc94ca044a5a14";
+    rev = "bcd25c8734edb5395ac239b16af336987da45e9d";
     fetchSubmodules = true;
-    hash = "sha256-44xABnopOfF8TilGQNVTynfRc0cUiiAk7E5KoRZ0Hw0=";
+    hash = "sha256-UzvJyyavnWwg7XoF1C5LbVaFRm5Ib3V7Sy3Tl/Rmomg=";
   };
 
   reisfmtSource = pkgs.fetchFromGitHub {
--- a/rtl/fpga/top_sonata.sv
+++ b/rtl/fpga/top_sonata.sv
@@ -196,7 +196,7 @@ module top_sonata
   import sonata_pkg::*;
 
   // System clock frequency.
-  parameter int unsigned SysClkFreq = 30_000_000;
+  parameter int unsigned SysClkFreq = 40_000_000;
   parameter int unsigned HRClkFreq  = 100_000_000;
 
   parameter SRAMInitFile    = "";
--- a/sw/cheri/CMakeLists.txt
+++ b/sw/cheri/CMakeLists.txt
@@ -7,7 +7,7 @@ include(FetchContent)
 
 FetchContent_Declare(CHERIOT_RTOS
   GIT_REPOSITORY    https://github.com/lowRISC/CHERIoT-RTOS
-  GIT_TAG           c29bdeccc0335b81acea09c069fc94ca044a5a14
+  GIT_TAG           bcd25c8734edb5395ac239b16af336987da45e9d
 )
 
 if(NOT DEFINED CMAKE_TOOLCHAIN_FILE)
--- a/sw/cheri/checks/revocation_test.cc
+++ b/sw/cheri/checks/revocation_test.cc
@@ -13,7 +13,7 @@
 
 using namespace CHERI;
 
-#define CPU_FREQ_HZ (30'000'000)
+#define CPU_FREQ_HZ (40'000'000)
 #include "../../common/defs.h"
 
 /**
--- a/sw/common/defs.h
+++ b/sw/common/defs.h
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#define CPU_TIMER_HZ (30'000'000)
+#define CPU_TIMER_HZ (40'000'000)
 #define BAUD_RATE (921'600)
 
 #define SRAM_ADDRESS (0x0010'0000)
--- a/sw/legacy/common/sonata_system.h
+++ b/sw/legacy/common/sonata_system.h
@@ -13,7 +13,7 @@
 #include "uart.h"
 
 // System Clock Frequency (Hz)
-#define SYSCLK_FREQ (30 * 1000 * 1000)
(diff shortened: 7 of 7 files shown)


elliotb-lowrisc:
I've tested it using Vivado v2021.1. I'd appreciate it if people running other Vivado versions could try building this and confirm it passes timing.

elliotb-lowrisc:
Correct post-route optimisation selection logic

elliotb-lowrisc:
Rebase and add hold fixing to mid-level post-route optimisation

elliotb-lowrisc:
Increase margin and rework post-route optimisation based on the results of further trials. Some post-route optimisation is now enabled by default, but it will not do anything unless it needs to.

elliotb-lowrisc:
Note that I have split out (and improved) the phys optimisations into a separate PR: https://github.com/lowRISC/sonata-system/pull/257

I will update this PR to be just the 40 MHz changes once the other is merged.

elliotb-lowrisc:
Rebase & remove separately-merged phys optimisations

elliotb-lowrisc:
Note that I believe the CHERIoT RTOS test suite's lack of support for different clock frequencies is the cause of the CI failures.

GregAC:
Did just try running this on top of latest main and unfortunately now seeing some timing failures, around -1.5ns WNS on clk_sys. We have just updated to the latest CHERIoT Ibex, it may have produced some trickier paths.

AlexJones0:
> Note that I believe the CHERIoT RTOS test suite's lack of support for different clock frequencies is the cause of the CI failures.

I wouldn't add this to your PR (instead, an upstream change to `cheriot-rtos` would be more appropriate), but you can use this patch to locally run against the `cheriot-rtos` test suite with `nix build .#tests-simulator -L` to check that it still works:

```patch
diff --git a/nix/software.nix b/nix/software.nix
index f4e3b11..8e7c7cd 100644
--- a/nix/software.nix
+++ b/nix/software.nix
@@ -39,7 +39,9 @@ in {
     buildInputs = cheriotPkgs ++ [lrPkgs.uf2conv];
     dontFixup = true;
     buildPhase = ''
-      xmake config -P ./tests/ --board=sonata-prerelease
+      cp ./sdk/boards/sonata-prerelease.json ./sdk/boards/sonata-40MHz.json
+      sed -i 's/    "timer_hz" : 30000000,/    "timer_hz" : 40000000,/' ./sdk/boards/sonata-40MHz.json
+      xmake config -P ./tests/ --board=sonata-40MHz
       xmake -P ./tests/
       ${../util/elf-to-uf2.sh} ./build/cheriot/cheriot/release/test-suite
     '';
```

Running this locally on this PR myself it seems to work for me.
Albeit my device is slower than CI, so I had to increase the timeout personally with the following patch:

```patch
diff --git a/nix/tests.nix b/nix/tests.nix
index 475cb45..f2ff32e 100644
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -41,7 +41,7 @@ in {
       python ${../util/test_runner.py} -t 60 sim \
           --elf-file ${sonata-system-software}/bin/test_runner
       echo "Test runner complete!"
-      python ${../util/test_runner.py} -t 600 sim \
+      python ${../util/test_runner.py} -t 1200 sim \
           --sim-boot-stub ${sonata-sim-boot-stub.out}/share/sim_boot_stub \
           --elf-file ${cheriot-rtos-test-suite}/share/test-suite
     '';
```

elliotb-lowrisc:
> Did just try running this on top of latest main and unfortunately now seeing some timing failures, around -1.5ns WNS on clk_sys. We have just updated to the latest CHERIoT Ibex, it may have produced some trickier paths.

Did you have [c248d55ae554a42e8](https://github.com/lowRISC/sonata-system/commit/c248d55ae554a42e8127f743036e12202cf2ae0f) in as well? Without it I get around -1.4 ns WNS, but with it the `phys_opt_design -directive RuntimeOptimized` command fixes timing (0.065 ns WNS).

GregAC:
> > Did just try running this on top of latest main and unfortunately now seeing some timing failures, around -1.5ns WNS on clk_sys. We have just updated to the latest CHERIoT Ibex, it may have produced some trickier paths.
> 
> Did you have [c248d55ae554a42e8](https://github.com/lowRISC/sonata-system/commit/c248d55ae554a42e8127f743036e12202cf2ae0f) in as well? Without it I get around -1.4 ns WNS, but with it the `phys_opt_design -directive RuntimeOptimized` command fixes timing (0.065 ns WNS).

I thought I had done. Anyway just tried rebasing this on latest main and it's meeting timing again at 0.223 WNS :)

So let's get this in, though obvious need to sort out the CHERIoT RTOS change in the board definition file.

GregAC (approved):
LGTM once we've sorted the change in CHERIoT RTOS (which needs a PR against https://github.com/lowRISC/cheriot-rtos/tree/sonata and then bumping the CHERIoT RTOS version in this repository for both cmake and nix.

GregAC:
> > LGTM once we've sorted the change in CHERIoT RTOS
> 
> @GregAC Do we just want to bump the `timer_hz` in the board description file to 40MHz, or do we want to look at making a small driver and reading this out of the system info block? The latter would require more changes to `sonata-software` but is probably more robust.

I'd certainly favour the latter. In particular it allows us to make future frequency bumps without disturbing software. Would be nice to keep push and see if we can hit 50 MHz but that's unlikely to happen for 1.0 for instance.

Main question is up-streaming the changes. Given it's a build time constant in the current setup could be a fairly big shift.

GregAC:
@AlexJones0 thinking about it a little more it really hinges on how quickly we can integrate the dynamic read to get frequency into CHERIoT RTOS. It feels like that's a reasonable task (1-2 days effort on the low end, could end up being far more) so may be worth just keeping it static for 1.0.

AlexJones0:
>  it really hinges on how quickly we can integrate the dynamic read to get frequency into CHERIoT RTOS

@GregAC  Looking at CHERIoT RTOS itself, it seems that they have checks that `CPU_TIMER_HZ` as created by the Xmake build is defined as a macro. The value is only used in `thread.h` to implement [`thread_microsecond_spin`](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/b9f206a7c9b138e762b51f539d085a4d7377a516/sdk/include/thread.h#L117), and it isn't being used for anything else. So I'm not sure if this is something upstream would even be willing to change. If they were I guess they wouldn't want to break existing compatibility, so I guess they would need to add a define like `CHERIOT_NO_USE_CPU_TIMER_HZ` which would then force you to use a version of the function that takes the frequency as an argument and disallows you from calling the current interface.

The alternative is requiring users to manually know that they shouldn't use this function or the value of `CPU_TIMER_HZ` and to instead use this dynamic read, but I'm not a huge fan of that as it seems quite error-prone and could cause a whole host of issues down the line.

Another hacky alternative would be setting `timer_hz` to e.g. 0 so that calls to the function fails, but that would provide unclear error messages and is reliant on upstream CHERIoT RTOS not adding more features that use `timer_hz` in the future. I'm not a huge fan of this approach either.

GregAC:
I think let's stay with the static frequency for now for simplicity.

Deciding how/if to support a dynamically determine frequency in upstream sounds like something that should be done but could well take some time and I'd rather not be maintaining a patch for this in our fork.

@HU90m / @marnovandermaas any thoughts?

elliotb-lowrisc:
Rebase and add rtos-bump commit

elliotb-lowrisc:
Thanks to @AlexJones0 for the RTOS changes

AlexJones0 (approved):
Software changes LGTM so since Greg approves & CI is passing now then this looks great, thanks @elliotb-lowrisc.

PR #240, opened 2024-10-14
SPI IRQ test

alees24:
Testing of SPI interrupts.

### DIFF
--- a/sw/cheri/tests/spi_tests.hh
+++ b/sw/cheri/tests/spi_tests.hh
@@ -269,6 +269,144 @@ int spi_flash_slow_clock_test(Capability<volatile SonataSpi> spi, ds::xoroshiro:
 }
 
 /**
+ * SPI controller interrupt test.
+ */
+int spi_irq_test(SpiPtr spi, ds::xoroshiro::P32R8 &prng, Log &log) {
+  // Ensure that the SPI SCLK speed is substantially slower than the system clock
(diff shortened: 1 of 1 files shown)


marnovandermaas (commented):
Hmmm, it looks like the pinmux test is failing after the interrupt test is added. Does the interrupt test leave the flash SPI in a different state than it started?

marnovandermaas:
I assume it's because of this line: `spi->control = SonataSpi::ControlInternalLoopback | SonataSpi::ControlTransmitEnable | SonataSpi::ControlReceiveEnable | (encRxWatermark << 8) | (encTxWatermark << 4);`
We might need to undo it after the interrupt test or set the correct control value in the pinmux SPI test.

alees24:
> I assume it's because of this line: `spi->control = SonataSpi::ControlInternalLoopback | SonataSpi::ControlTransmitEnable | SonataSpi::ControlReceiveEnable | (encRxWatermark << 8) | (encTxWatermark << 4);` We might need to undo it after the interrupt test or set the correct control value in the pinmux SPI test.

This isn't a problem with the interrupt test. Altered signalling to the SPI DPI model in simulation with the pinmux changes has exposed the fact that the model was not happy with SCK toggling when CS is deasserted (PR #292).

marnovandermaas (approved):
Thanks for fixing this. I think this is good to have in CI.

PR #241, opened 2024-10-14
Correct tx watermark check

alees24:
Typo in tx watermark logic; was using rx config in equality test.
Let's at least make sure that the silly typo is fixed in time.


AlexJones0 (approved):
LGTM

PR #252, opened 2024-10-16
Error LEDs fixed

marnovandermaas:
This removes PermitStoreLocalCap from error LEDs since this is no longer supported in CHERIoT Ibex. It also make the cycling on FPGA go faster which is a nicer effect.


alees24 (approved):
Looks good. Minor point: there's a typo in the second commit message.

marnovandermaas:
Thanks, I've updated the commit message!

PR #257, opened 2024-10-17
Add phys-based timing optimisations

elliotb-lowrisc:
Add flow hooks to implement some optional timing optimisations:
- Add timing margin to clk_sys for early implementation stages. This allows faster clk_sys speeds w/o much additional runtime.
- Add optional additional logic optimisation.
- Add two levels of post-route optimisation commands. This allows even faster clk_sys speeds at the cost of added runtime.

All these optimisations are controlled by an `opt` TCL variable set in the hook that is first called in the implementation flow.


GregAC (approved):
Thanks @elliotb-lowrisc nice to see continuing improvements in the FPGA flow.

PR #259, opened 2024-10-17
Update OpenTitan IP blocks

alees24:
This PR updates the vendored-in IP blocks to the versions used in Earl Grey production, updates the vendor patch files accordingly, and makes the necessary RTL changes to accommodate altered ports and interrupts.

- tlul_adapter_sram has acquired some additional ports for the readback checking functionality, which is not required in Sonata.
- I2C and UART interrupts have been modified slightly.
- The newer I2C block has separated controller and target functionality and offers multi-controller support. This change leads the controller logic to return to an Idle state upon detecting SDA/SCL interference, so the controller must see its own outputs and it no longer suffices to model the I2C bus as two separated unidirectional channels; combine the outbound and inbound signalling within the Verilator test bench.

`tests/test_runner` passes in Verilator simulation with all devices/tests, and the FPGA build is timing clean; local testing has been performed on FPGA except for those I2C devices that are unavailable.


vendor/patches/lowrisc_ip/reggen/0001-Remove-Integrity-From-Reg-Top.patch
marnovandermaas, line 1: Why did you remove this patch? is this integrity stuff no longer in the template? Have you tested this by running `util/reg_gen.sh`?

vendor/lowrisc_ip/util/reggen/reg_top.sv.tpl
```diff
@@ -254,7 +310,7 @@ module ${mod_name} (
     ## passthrough, we must generate data integrity for it here.
   tlul_rsp_intg_gen #(
     .EnableRspIntgGen(0),
-    .EnableDataIntgGen(0)
+    .EnableDataIntgGen(1)
```
marnovandermaas, line 313: I don't think we want these changes.

vendor/lowrisc_ip/util/reggen/reg_top.sv.tpl
```diff
@@ -212,11 +214,65 @@ module ${mod_name} (
...
+  assign intg_err_o = 1'b0;
+% endif
+
   // outgoing integrity generation
   tlul_pkg::tl_d2h_t tl_o_pre;
   tlul_rsp_intg_gen #(
-    .EnableRspIntgGen(0),
-    .EnableDataIntgGen(0)
+    .EnableRspIntgGen(1),
+    .EnableDataIntgGen(${common_data_intg_gen})
```
marnovandermaas, line 275: I don't think we want these changes.

rtl/system/sonata_system.sv
```diff
@@ -1000,9 +1007,12 @@ module sonata_system
       .tl_o                 (tl_uart_d2h[i]),
 
       // Interrupts.
+      // Note: the indexes here match the bits in the `intr_` registers,
+      // but we also keep the port ordering the same as the module.
       .intr_tx_watermark_o  (uart_interrupts[i][0]),
+      .intr_tx_empty_o      (uart_interrupts[i][8]),  // Interrupt was appended.
       .intr_rx_watermark_o  (uart_interrupts[i][1]),
-      .intr_tx_empty_o      (uart_interrupts[i][2]),
+      .intr_tx_done_o       (uart_interrupts[i][2]),
```
marnovandermaas, line 1015: Why did you not leave empty at index 2 and append done at 8?

marnovandermaas (commented):
Thanks for doing this import (I know it's a lot of fiddly work). Just a few comments from my side before we can merge this. Does this PR come with any RTOS driver visible changes?

rtl/system/sonata_system.sv
GregAC: Given this isn't software visible, we're just ORing it all together, does it matter?

HU90m (approved):
LGTM

vendor/patches/lowrisc_ip/rv_plic/0001-Remove-Alerts.patch
```diff
@@ -1,122 +0,0 @@
-diff --git a/data/rv_plic.hjson.tpl b/data/rv_plic.hjson.tpl
```
GregAC, line 1: Don't we still need this? As part of this PR we should re-generate the PLIC with the updated templates.

rtl/system/sonata_system.sv
alees24: I've annotated it above the connections; the more elegant solution would be to reorder the output ports in `uart.sv` but I didn't really want to do that. My concern in shuffling the indexes of `uart_interrupts` was simply that somebody expanding it in a waveform viewer may be misled by assuming that the indexes match the interrupt numbers.

vendor/patches/lowrisc_ip/reggen/0001-Remove-Integrity-From-Reg-Top.patch
alees24: The reggen stuff wasn't in the repo when I started; rebased now.

vendor/patches/lowrisc_ip/rv_plic/0001-Remove-Alerts.patch
alees24: Rebased and regenerated PLIC now; I've opted to leave reggen/templates such that it includes the `OpenTitan project` reference and so this is present in the generated PLIC, but left the Sonata-specific RTL (which I haven't needed to update) without reference in the copyright headers; this does mean - for now - that the output from util/reg_gen.sh mismatches trivially, but it sounds as though there are other twiddles to be done also at a later date.

Issue #124, opened 2024-06-14
Documentation problems


benlaurie:
Firstly, section 1.3 (https://lowrisc.org/sonata-system/doc/guide/building-examples.html) actually has section 2.2 (https://lowrisc.org/sonata-system/doc/dev/toolchain-setup.html) as a prerequisite - this is highly confusing, 2.2 should come first.

Secondly `xmake config --board=sonata` is incorrect, it also needs the location of the sdk specified: `xmake config --sdk=$LLVM_PATH/builds/cheriot-llvm --board=sonata` ought to work if 2.2 has just been done. :-)

Thirdly, the uf2 converter is a prerequisite that is not documented: https://github.com/makerdiary/uf2utils?tab=readme-ov-file

Issue #97, opened 2024-05-24
xmake is not listed as a prerequistie


benlaurie:

marnovandermaas:
Thanks Ben! I recently split the software toolchain setup page into a Nix and a manual setup. The Nix setup should point you to the sonata software repository and `xmake` is included in the Nix environment: https://lowrisc.org/sonata-system/doc/guide/toolchain-setup.html?highlight=requi#nix-setup

benlaurie:
OK, but it isn't included in the manual setup.

davidchisnall:
Please can you point people at the CHERIoT docs, rather than trying to duplicate half of them?

marnovandermaas:
Ok, I've listed xmake as a dependency for Sonata software and have linked to the CHERIoT RTOS documentation for a full development setup: https://github.com/lowRISC/sonata-system/pull/260

PR #260, opened 2024-10-18
Doc update to close some old issues

marnovandermaas:
Fixing up two lingering documentation issues.

Closes: https://github.com/lowRISC/sonata-system/issues/124
Closes: https://github.com/lowRISC/sonata-system/issues/97

### DIFF
--- a/doc/dev/toolchain-setup.md
+++ b/doc/dev/toolchain-setup.md
@@ -79,3 +79,8 @@ Note the checkout and build will take some time, and the build process may have 
 This should put the binaries in a `bin` subdirectory of your build folder (the `builds/cheriot-llvm` folder you made).
 
 With the CHERIoT/LLVM toolchain built, you can now continue to setup the examples and build them.
+
+### Additional Sonata Software dependencies
+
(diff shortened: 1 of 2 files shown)


doc/dev/toolchain-setup.md
```diff
@@ -79,3 +79,8 @@ Note the checkout and build will take some time, and the build process may have
 This should put the binaries in a `bin` subdirectory of your build folder (the `builds/cheriot-llvm` folder you made).
 
 With the CHERIoT/LLVM toolchain built, you can now continue to setup the examples and build them.
+
+### Additional Sonata Software dependencies
+
+To set-up a manual environment that can build Sonata Software and RTOS you can follow the [building dependencies section of the RTOS guide](https://github.com/CHERIoT-Platform/cheriot-rtos/blob/main/docs/GettingStarted.md#building-the-dependencies).
```
AlexJones0, line 85: Grammar nit: "set-up" is an alternative form of the noun form of "setup", whereas here it should be the verb form "set up".

Also might be worth writing "... Software and **CHERIoT** RTOS ..." just to remove doubt about which RTOS you mean, since I don't think it's mentioned elsewhere in this markdown file.

AlexJones0 (approved):
Thanks for looking at these issues, LGTM - just one minor nit.

doc/dev/toolchain-setup.md
marnovandermaas: Good point, these should now be changed.

PR #264, opened 2024-10-22
Block diagram includes pinmux and larger text

marnovandermaas:
Add pinmux and padring to block diagram.


HU90m (changes requested):
thanks for updating

1) I don't think the ADC will be passed through the pinmux
2) I did/does look like the user USB sits between the debug module and the bus. We probably have enough space that we don't have to have blocks sitting on atop buses.
3) While you're moving the user USB, it would be nice to have wire leaving the FPGA from it.

*change request just for list item 1.*

marnovandermaas:
There's actually more issues with this block diagram. Closing this for now.

PR #267, opened 2024-10-23
Introduce loopbacks for testing in simulation

alees24:
Loopback functionality for SPI, GPIO and UART connections so that the FPGA tests will operate in Verilator simulation too.
Delete a legacy TODO comment, now that we actually model more I/O functionality.


AlexJones0 (approved):
LGTM as long as CI passes, thanks for doing this @alees24

PR #271, opened 2024-10-24
Pin Mappings

HU90m:
I think this is a sensible pin mapping for V1.0, feedback appreciated.

I wrote them up in [a dot file](https://github.com/HU90m/sonata-system/blob/a25f156df8e6a5ccb1af1f1f61d719b6fe1b0ec1/doc/ip/pinmux/pin_mappings.dot#L311-L525), because it has all the semantics I needed, makes editing easier and renders to [a nice diagram](https://github.com/HU90m/sonata-system/blob/doc/doc/ip/pinmux/pin-mappings.md).

A brief explanation:
1. It is likely users won't want a single SPI block/compartment handling both Ethernet and GPIO. So SPI0 is reserved for the devices that are likely to be more security critical. The downside is that if someone wants to download onto flash or the sd card, there'll be contention on the SPI block.
2. I've used the SPIs CS lines for LCD reset & command/data lines and Ethernet reset, so that the device can be properly managed by one compartment.
3. The Hats share as many blocks as possible because it is larger hats block port access for smaller hats. Pmod on the other hand is a little more independent.

Points to consider:
1. There are only two UARTs, we may want to bump this to three.
2. The Pmod and Hat GPIO could be collapsed into one. Leaving only two GPIO blocks.
3. This assumes we'll create an additional 5 PWMs
4. The gpio output enable signals are useless on in the general purpose GPIO
5. There is no rs485 connected up at the moment.


GregAC:
Thanks for this @HU90m, big fan of the dot file, we should look at getting that auto-generated but certainly not a priority now.

A few comments:

- Agreed on the sharing of Flash/Ethernet/SD card on SPI0. I can see an issue where you want high performance access to all of these so you'd prefer dedicated SPI but we can always add more SPI if this is raised as an issue. I figure we can have the bootloader use the flash and then switch the pinmux to use ethernet just before it does the jump into the binary. Then our base setup is much like we had before (where only the ethernet was getting used by CHERIoT RTOS). There's a system architecture question here of how one can safely share that SPI amongst different compartments. Perhaps one compartment holds the SPI capability and can provide it to other compartments via a cross compartment call but the capability received doesn't have the global permission set. The same compartment that holds the SPI capability would also have access to the relevant pinmux registers so it would set the pinmux up and then provide the capability. However we don't need to worry about that right now.

- There's no need to shrink GPIO blocks, as they're not expensive, instead prefer to have separate GPIO blocks for separate headers/usages for ease of use and isolation. I would suggest separate GPIO registers in the following groupings
  - RPI Hat pins
  - Mikrobus pins
  - Arduino pins
  - PMOD0
  - PMOD1
  - PMODC
  - User LEDs and buttons/switches
  - RS485 REn/DE pins
  If I've understood the dot file correctly you've proposed muxing one GPIO block to multiple possible pins. There should be no need to do this. A particular bit in a particular GPIO block should only ever mux to one particular pin.

- The capabilities of the two pmods should be identical, i.e. they both support 1A, 2A, 3A and 6A

- We should have one extra UART and SPI. I think the LCD should get a dedicated SPI (as I can imagine many things wanted to use in conjunction with other SPI things), perhaps in the following configuration:
  - SPI0: Flash, Ethernet, SD Card
  - SPI1: LCD (with the CS lines used to connect up the various extra IO here, with the exception of the backlight, see below)
  - SPI2: RPI SPI 0, Arduino Header, PMOD 0, Mikrobus
  - SPI3: RPI SPI 1, Arduino ICSP header, PMOD 1
  - UART0: USB Ser 0
  - UART1: USB Ser 1, RPI UART, Arduino UART, Mikrobus UART, PMOD0
  - UART2: USB Ser 1, RS232, RS485, PMOD 1

 I've mapped USB Ser 1 to both UART1/UART2 here this is to give some flexibility so if you want two USB UART channels you can have that along with some other serial without constraint.

- Add an extra PWM channel to use for the LCD backlight

SUMMARY.md
```diff
@@ -23,7 +23,7 @@
     - [PWM](doc/ip/pwm.md)
     - [HyperRAM](doc/ip/ram.md)
     - [ADC](doc/ip/adc.md)
-    - [Pinmux](doc/ip/pinmux.md)
+    - [Pinmux](doc/ip/README.md)
```
marnovandermaas: Is this correct? Should this be doc/ip/pinmux/README.md?

util/top_gen/generator.py
```diff
@@ -365,7 +365,7 @@ def legacy_ordering(input_pin: OutputBlockIo) -> tuple[int, int, int]:
         ),
         ("rtl/templates/sonata_pkg.sv.tpl", "rtl/system/sonata_pkg.sv"),
         ("rtl/templates/pinmux.sv.tpl", "rtl/system/pinmux.sv"),
-        ("doc/ip/pinmux.md.tpl", "doc/ip/pinmux.md"),
+        ("util/templates/pinmux.md.tpl", "doc/ip/pinmux/README.md"),
```
marnovandermaas, line 368: I think I would prefer having this in doc/templates/pinmux.md.tpl

marnovandermaas (commented):
I would prefer to have one more SPI and a different configuration:
- 1 dedicated SPI for the LCD
- 1 dedicated SPI for Ethernet
- 1 SPI for flash/microSD, one Arduino or one Pi
- 1 SPI for one Arduino, Shield and one PI

I think the dedicated SPI for LCD and Ethernet makes sense. I also think that it would be nice for applications to be able to use both Flash and Ethernet.

HU90m:
Updated to match suggestions in https://github.com/lowRISC/sonata-system/pull/271#issuecomment-2436776664 .

With 2 exceptions:
- removed arduino ICSP. *We don't have a spi device block.*
- still no RS485. *This can come in a future release.*

GregAC:
> still no RS485. This can come in a future release.

I'll take a look at adding this in, would be nice to have for 1.0. I'll aim to build on top of the work you do so won't add to your workload.

HU90m:
> > still no RS485. This can come in a future release.
> 
> I'll take a look at adding this in, would be nice to have for 1.0. I'll aim to build on top of the work you do so won't add to your workload.

That would be amazing, thank you!

SUMMARY.md
marnovandermaas: This is now resolved.

util/top_gen/generator.py
marnovandermaas: Leaving this in util for now so that the doc directory contains only readable files.

marnovandermaas (approved):
This PR is good to go.

GregAC:
Thanks for the changes @HU90m, one question: shouldn't we have a dedicated PWM channel for the LCD backlight? As is you'll need to turn off the backlight if you want to use the PWM it shares with the HAT or shield.

HU90m:
> Thanks for the changes @HU90m, one question: shouldn't we have a dedicated PWM channel for the LCD backlight? As is you'll need to turn off the backlight if you want to use the PWM it shares with the HAT or shield.

One can use the LCD SPI's 4th CS as an alternative. Is that sufficient flexibility?

GregAC:
It just feels odd, PWM channels aren't hugely expensive and saying you can only run the LCD on full brightness if you want to use the full PWM set in the shield or the hat is an odd restriction.

HU90m:
> It just feels odd, PWM channels aren't hugely expensive and saying you can only run the LCD on full brightness if you want to use the full PWM set in the shield or the hat is an odd restriction.

I'll add another PWM for the LCD. Unless @marnovandermaas has any opinions?

HU90m:
> Thanks sorry to be a pain but I would like to avoid weird things in the v1.0 release if possible as we could be stuck with them for quite some time!

Don't be! I'm completely agree

PR #272, opened 2024-10-24
Expand both checking and logging of IRQs

alees24:
Present logging information about the IP-level interrupt state, since this is more useful in the event of a failure.

Extend the IP-level interrupt state testing for SPI and USBDEV since this was incomplete.


engdoreis (approved):

engdoreis:
You may need to increase the timeout for the FPGA test as well.

alees24:
@AlexJones0, @engdoreis The tests, although completing in mere seconds, have been timing out purely because of the `tick.sleep()` per output line; I see no reason for that to be there anyway. Please check my reasoning.

PR #277, opened 2024-10-25
Update Verilator top level and I2C DPI models to match Sonata board

alees24:
The two DPI models did not model the post-pinmux changes to the I2C buses; there are no longer just two physical I2C buses electrically commoned within the FPGA.

Separate the I2C devices into three I2C DPI models representing the physical buses that have devices at present. Tie the other I2C buses inactive, modelling the pullups on the FPGA board.

Modify `i2cdpi` so that it properly handles traffic to an address at which there is no connected device.


AlexJones0 (approved):
Thanks @alees24, this appears to be working with the I2C Pinmux test now. The changes seem sensible so I'm going to merge this so that I can get the pinmux test PRs open. If there are any problems, we can iterate with further PRs. Would still be good to get @marnovandermaas's comments even after this is merged though.

PR #280, opened 2024-10-28
[doc] Tweak release proceedure

GregAC:
Note that a detailed timing report should be run where timing failures are reported as Vivado sometimes reports negative slack that disappears with a detailed timing report


marnovandermaas (commented):
I'm not sure I want that in the release procedure to be honest. I prefer to understand why this is happening and removing the constraints if necessary.

GregAC:
As I said in the other PR it's likely a tooling bug, so worth noting it can be a problem so we're not gating releases on imaginary timing issues.

PR #283, opened 2024-10-28
Recover from low Rx input. Reduce test duration by shortening time delays and using higher I2C speed.

alees24:
Prevent a low GPIO input to the UART Rx input causing the test to run indefinitely without timeout. A low Rx input keeps the UART RX logic active and prevents it becoming Idle except for infrequent single-cycle assertions of Idle.
Also switch the output driver before redirecting the input for more predictable behaviour.

Use higher I2C speed when reading the RPi HAT ID EEPROM to reduce the test duration substantially.

Reduce the timeout values in the UART and GPIO testing to more practical values for FPGA and reduce the simulation time.


AlexJones0 (approved):
LGTM, thanks @alees24!

PR #284, opened 2024-10-28
Automatically generate the pinmux driver from `top_config.toml`

AlexJones0:
This PR modifies the Pinmux driver `platform-pinmux.hh` so that it is now automatically generated by `top_gen.py`, i.e. by running `py util/top_gen.py`. This is done using a corresponding template file `sw/cheri/common/platform-pinmux.hh.tpl`. The output of the driver is designed to match the previous output (see changes to `platform-pinmux.hh`), apart from a few key aspects:
 - After discussion with @HU90m and @marnovandermaas, I've switched from using driver-specific pin/block names to matching the "doc"/pin names used by `top_gen` already. This minimises the amount of work that this PR is doing and makes it easier to use for development. This could change again in the future when it comes to upstreaming the finalised driver to CHERIoT RTOS to meet coding conventions, but this is being kept simple for now.
 - As such, the above use of pin names means that the comments mapping the pin names are no longer needed. We also decided to omit the locations, so that this PR barely changes `top_gen` at all, and so that this is instead handled by appropriate documentation.
 - Other than that, the output is entirely the same, apart from the switch cases in the `block_input_options` functions which were already wrong because of an accident I made when rebasing previously - `spi_3_rx` should have 5 muxable options and not 4 based on the current `pinmux.md`. This issue appeared because before rebasing the manual check PR I gave `spi_3_rx` 4 options, but after rebasing the MicroSD was added, so I needed to bump this to 5 (but I missed this). This is a good example of why having this file auto-generated is helpful.

See commit messages for more details. This has been tested with the existing pinmux tests (which will also run in CI), as well as the manual pinmux check procedure (see #230) and the small manual pinmux check over UART0. All pass as expected.

### DIFF
--- a/sw/cheri/checks/pinmux_all_blocks_check.cc
+++ b/sw/cheri/checks/pinmux_all_blocks_check.cc
@@ -77,33 +77,33 @@ using namespace CHERI;
   // initializers, as that requires `memcpy` to exist which we do not have.
   // Likewise, arrays with multiple pins/inputs must be individually set, or
   // we will get errors with `memcpy`.
-  OutputPinAssignment pmod_test_gpio_on_pins[]     = {{SonataPinmux::OutputPin::Pmod0Io2, 1}};
-  BlockInputAssignment pmod_test_gpio_on_inputs[]  = {{SonataPinmux::BlockInput::PmodGpio2, 1}};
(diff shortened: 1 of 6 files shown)


marnovandermaas (approved):
This looks good to me, thanks for making this change. Much nicer than having to update it manually.

PR #287, opened 2024-10-29
pmod: changed pin names to match the schematic names

HU90m:


AlexJones0:
Force push has rebased the PR on latest `main`, and has made the software changes to the manual "all blocks" pinmux check required to make this pass CI. I have also personally tested the updated manual pinmux check on top of a local bitstream build of the PR and it still appears to be working fine (when using the changes from #288 on top, as this test is currently broken in `main`) with these name changes, as we would expect.

AlexJones0:
Latest push rebases on `main` (after the fix from #288 was added), and adds more detailed commit messages to the two commits that I added.

marnovandermaas (approved):
Looks good to me good to correct the PMOD Names.

PR #262, opened 2024-10-21
Now passing SPI chip selects through pinmux

HU90m:
*see commit messages*

PR #288, opened 2024-10-29
Fix "all blocks" manual pinmux check so it works again after Chip Select added to Pinmux

AlexJones0:
This PR fixes the "all blocks" manual pinmux check (and the pinmux checker in general) so that the latest `main` which includes changes from #262 will now work again. See the commit messages for more details.

This PR primarily does 3 things:
1. Updates the Pinmux config to add the SPI Chip Select to PMOD, where it was previously missing. Whilst at it, I also updated PMOD0 to use SPI2 instead of SPI3, and PMOD1 to use SPI3 instead of SPI4, because it appears that the eventual pinmux mapping (see [here](https://github.com/lowRISC/sonata-system/blob/8df6c195198b7b0121173351391f841b0896a176/doc/ip/pinmux/pin_mappings.svg)) intends to use this mapping. By making this smaller change now in this PR, it reduces the amount of work we have to do later, as the "all blocks" test should stay passing and not break with that PR.
2. It then updates the pinmux checker and the "all blocks" check to use the above features. That is, it modifies the SPI test in the pinmux checker to now use the SPI's CS rather than just a GPIO pin, and it changes the available SPI to be 2-3 instead of 3-4 as before. It updates the pinmux mappings and SPI instance being used in the "all blocks" testplan to match the new values in `pinmux.md` after the config change.
3. Finally, it fixes a couple of minor issues I found while looking at the code. One is just a small good practice (use `sizeof(jedec_id)` instead of `3`), and the other is changing the UART test so that it now actually uses the `timeout` and `test_length` values provided in the test, rather than hard-coded literals as I had apparently accidentally done previously.

This has been tested with the existing pinmux tests (which will also run in CI), as well as the manual "all blocks" check itself, and the small manual pinmux check over UART0. All pass as expected.

### DIFF
--- a/data/top_config.toml
+++ b/data/top_config.toml
@@ -423,14 +423,15 @@ block_ios = [{block = "pwm", instance = 0, io = "ios", io_index = 0}]
 [[pins]]
 name = "pmod0_0"
 block_ios = [
-  {block = "gpio", instance = 2, io = "ios", io_index = 0},  # can be used as SPI CS
+  {block = "gpio", instance = 2, io = "ios", io_index = 0},
+  {block = "spi", instance = 2, io = "cs", io_index = 0},
  ]
 
 [[pins]]
 name = "pmod0_1"
 block_ios = [
   {block = "gpio", instance = 2, io = "ios", io_index = 1},
-  {block = "spi", instance = 3, io = "tx"},
+  {block = "spi", instance = 2, io = "tx"},
   {block = "uart", instance = 2, io = "tx"},
 ]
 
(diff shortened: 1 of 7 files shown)


AlexJones0:
Force push is rebasing on `main` to address merge conflicts with the PWM mapping changes that just went in.

AlexJones0:
Have force pushed to add 9b84346f759980ed314c0174f8c52bb7f7d8516d to hopefully make the code a bit cleaner. No other changes. From manual testing on the previous bitstream everything still seems to pass fine.

alees24 (approved):
Thanks for the tidy ups. Let's just be sure that CI is happy.

AlexJones0:
CI is all passing so I'm going to merge, thanks for the review @alees24.

PR #291, opened 2024-10-29
Improve Pinmux testing to account for SPI Chip Select

AlexJones0:
See commit messages for more details.

Alongside a couple of code improvements, this PR primarily focuses on the following:
 - Make the existing SPI Pinmux test also mux the Flash's CS pin.
 - Make the SPI Pinmux test more resilient by resetting the flash after muxing any SPI pins, to recover from erroneous states caused by muxing pins.
 - Extend the SPI Pinmux test to specifically check whether muxing of the CS is working. That is, it only disables the CS via pinmux and checks that the test fails, and then re-enables it and checks that it passes again. This is to give confidence that the pinmux surrounding chip selects specifically is working, as this is an area that has undergone a lot of changes recently in the RTL.

This has been tested to pass on FPGA locally; as it only touches the tests nothing should change with regards to other pinmux checks.

### DIFF
--- a/sw/cheri/tests/pinmux_tests.hh
+++ b/sw/cheri/tests/pinmux_tests.hh
@@ -123,12 +123,14 @@ static int pinmux_uart_test(SonataPinmux *pinmux, ds::xoroshiro::P32R8 &prng, Ua
 static int pinmux_spi_flash_test(SonataPinmux *pinmux, Capability<volatile SonataSpi> spi, SpiFlash spi_flash) {
   constexpr uint8_t PmxSpiFlashDataToSpiTx0   = 1;
   constexpr uint8_t PmxSpiFlashClockToSpiClk0 = 1;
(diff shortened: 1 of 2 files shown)


AlexJones0:
Currently blocked on a fix for the fail in simulation, due to the fact that chip select pinmux config should default to `1` when de-selected and not `0`.

AlexJones0:
Force push was a rebase on `main`, primarily to pull in the commit from #296 so that this test should now pass on simulator as well.

AlexJones0:
PR is probably still failing because the change in #296 was not also replicated in `top_verilator.sv`.

AlexJones0:
Force push is a rebase on `main`, to now pull in the commit from #300 so that this test should *now* hopefully pass on simulator as well :sweat_smile:

alees24 (approved):
LGTM, thanks. I'm happy for this to be merged when it (finally, following all the churn) goes through CI.

PR #292, opened 2024-10-29
Protect SPI DPI model reads against SCK toggling without CS assertion

alees24:
Driving out the clock to a simulated SPI device without corresponding assertion of CS caused the bit-level handling to get out of step, resulting in a failure to read JEDEC ID from the SPI flash after (in this case) the IRQ testing performs a small number of byte reads with the CS line intentionally deasserted (internal loopback mode).


AlexJones0 (approved):
Looks good to me, thanks for this change @alees24.

PR #290, opened 2024-10-29
gpio: Split the gpio into separate blocks

HU90m:
The last of the mammoths

PR #293, opened 2024-10-30
All SPI CS enables must be pulled high

alees24:
The SPI controller does not produce enable signals for the output CS lines, they are a requirement of pinmux. Ensure that all are pulled high, and not just the first.

This would impact other SPI devices as and when they are connected, but prevented operation of the LCD module in #290.


marnovandermaas (approved):
This looks right to me.

PR #294, opened 2024-10-30
Provisional update to lcd_check for pinmux/GPIO changes

alees24:
This optionally switches the lcd_check code to drive the LCD signals via the SPI controller itself rather than the GPIO signals used previously.

~~ To help with testing/ratification of #290 ~~
Converting to draft because the lcd connectivity changes were reverted at some point in that PR, and the additional LCD pins are still driven by GPIO presently.


alees24:
Superseded.

PR #297, opened 2024-10-30
doc: Updated pinmapping diagram

HU90m:
1) The previous mapping had the pmod sda and scl pins the wrong way around.
2) The `selsw` has been added, previously forgotten.
3) mikroBUS doesn't specify GPIO functionality, so the gpio_mb has been removed.


doc/ip/pinmux/pin_mappings.dot
```diff
@@ -438,7 +445,8 @@ digraph {
   pwm:2 -> pmod0:6;
   pwm:3 -> pmod1:2;
   pwm:4 -> pmod1:6;
-  pwm:5 -> lcd:backlight;
```
marnovandermaas, line 441: what's connected to pwm 5?

marnovandermaas (approved):
Looks reasonable to me just one question.

doc/ip/pinmux/pin_mappings.dot
HU90m: `rph:g21_spi1_sclk` and `ah:tmpio11`

marnovandermaas: Great thanks!

PR #300, opened 2024-10-30
Improve simulation match with FPGA

alees24:
Re-express some wiring and ensure that inputs are defined within top_verilator to achieve a better match with the FPGA implementation in top_sonata.

In particular ensure that the output pins are driven high when disabled as per the the FPGA builds, which is important for SPI CS and UART TX lines.

### DIFF
--- a/dv/verilator/top_verilator.sv
+++ b/dv/verilator/top_verilator.sv
@@ -232,6 +232,14 @@ module top_verilator (input logic clk_i, rst_ni);
   assign rph_g16_ce2  = inout_to_pins[INOUT_PIN_RPH_G16_CE2];
   assign mb1          = out_to_pins[OUT_PIN_MB1];
 
+  logic unused_out_pins = ^{out_to_pins[OUT_PIN_RS232_TX],
+                            out_to_pins[OUT_PIN_ETHMAC_COPI],
+                            out_to_pins[OUT_PIN_ETHMAC_SCLK],
+                            out_to_pins[OUT_PIN_MB2],
+                            out_to_pins[OUT_PIN_MB10],
+                            out_to_pins[OUT_PIN_MICROSD_CMD],
+                            out_to_pins[OUT_PIN_MICROSD_CLK]};
+
   logic [18:0] unused_gp_o;
 
   // Loopback functionality used to verify the operation of the pinmux and GPIO pins;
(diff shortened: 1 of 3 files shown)


dv/verilator/top_verilator.sv
```diff
@@ -232,6 +232,13 @@ module top_verilator (input logic clk_i, rst_ni);
   assign rph_g16_ce2  = inout_to_pins[INOUT_PIN_RPH_G16_CE2];
   assign mb1          = out_to_pins[OUT_PIN_MB1];
 
+  logic unused_out_pins = ^{out_to_pins[OUT_PIN_ETHMAC_SCLK],
+                            out_to_pins[OUT_PIN_ETHMAC_CS],
+                            out_to_pins[OUT_PIN_MB2],
+                            out_to_pins[OUT_PIN_MB10],
+                            out_to_pins[OUT_PIN_MICROSD_CMD],
+                            out_to_pins[OUT_PIN_MICROSD_CLK]};
```
AlexJones0, line 240: I think you might be missing a couple of pins like `OUT_PIN_RS232_TX` and `OUT_PIN_ETHMAC_COPI` here?

AlexJones0 (approved):
Looks good to me, thanks @alees24.
Might be judicious to get someone who works on hardware to review this even after it is merged, but all these RTL changes look good from my understanding if CI is still passing.

rtl/system/sonata_system.sv
```diff
@@ -1205,15 +1211,15 @@ module sonata_system
     .spi_sck_i(spi_sck),
     .spi_sck_en_i('{default: 'b1}),
     .spi_cs_i(spi_cs),
-    .spi_cs_en_i('{default: {SPI_CS_WIDTH{1'b1}}}),
+    .spi_cs_en_i('{default: '1}), // All continuously enabled.
```
marnovandermaas, line 1214: I thought we needed this change to make them all 1?

dv/verilator/top_verilator.sv
alees24: Oops, yes, thanks.

rtl/system/sonata_system.sv
AlexJones0: My (admittedly very limited) understanding of SystemVerilog was that using `'1` instead of `1b1` will automatically extend to fill the correct size, so isn't this change essentially doing the same thing? (as `spi_cs_en_i` should match the size of `SPI_CS_WIDTH`?) So it's just a refactor to clean up?

alees24: It was previously expressed as `1'b1` which sets only the LSB because it's a sized literal; this replaces it with the (arguably) more elegant unsized, single-bit literal that I think was originally intended, and which replicates the bit. (5.7.1 Integer literal constants).
I have of course checked that they do indeed all get set to 1.

PR #301, opened 2024-10-31
Spit xbar into two levels

elliotb-lowrisc:
Attempt to reduce the depth of xbar logic on critical timing paths to/from the HyperRAM/SRAM by moving most peripherals from the main cross-bar to a new "peri" (peripheral) sub-crossbar. Keeps addresses the same, but some few devices have increased access latency due to being moved to xbar_peri or to improve QoR.

Blocks with increased access latency: timer, system_info, & hw_rev.

Seems to improve timing and area utilisation.


elliotb-lowrisc:
rebased onto draft pinmux changes

elliotb-lowrisc:
rebased

rtl/templates/sonata_xbar_main.sv.tpl
```diff
@@ -93,24 +131,14 @@ module sonata_xbar_main
...
     .tl_timer_i       (tl_timer_i),
-    .tl_spi_board_i   (tl_spi_board_i),
-    .tl_spi_board_o   (tl_spi_board_o),
-    .tl_spi_lcd_i     (tl_spi_lcd_i),
-    .tl_spi_lcd_o     (tl_spi_lcd_o),
-    .tl_spi_ethmac_i  (tl_spi_ethmac_i),
-    .tl_spi_ethmac_o  (tl_spi_ethmac_o),
     % for block in config.blocks:
-    % if not block.name == "gpio":
+    % if block.name not in ["gpio", "i2c", "spi"]:
```
HU90m, line 135: I think it may be nicer to add field to a block which states which crossbar it should be in. Instead of hard coding certain blocks in the templates.

HU90m (changes requested):

rtl/templates/sonata_xbar_main.sv.tpl
elliotb-lowrisc: Fair enough, I've had a go. How does it look to you now?

elliotb-lowrisc:
rebase

PR #303, opened 2024-10-31
Relax timing constraints on some SPI I/O

elliotb-lowrisc:
Getting tight on timing on some SPI I/O paths, so relax some of the more arbitrary ones to help overall timing pressure.


GregAC (approved):
Thanks @elliotb-lowrisc

PR #305, opened 2024-10-31
LCD DPI model and lcd_check updates

alees24:
Modify the use of GPIO pins to match hardware changes.

The DPI model of the LCD panel now supports all rendering orientations for other tests in simulation.
Modify the lcd_check code to support operation in landscape orientation and not just portait. This was necessary to assist with bring up of other simulation tests; 3 of the possible 8 orientations are in use.


elliotb-lowrisc (approved):
Works when I test it with a current bitstream build

marnovandermaas (approved):
Test works for me, both landscape and portrait. I haven't checked the DPI part of it.

PR #306, opened 2024-10-31
Pinmux no default output

GregAC:
First commit is just a tweak to the comments in the pinmux template (and hence effects the generated RTL). The second commit adds a feature to the pinmux generation, which does not alter the generated RTL. I will use this feature for the RS485 integration and we should likely it apply to other pins as well (otherwise things like UART and SPI outputs will appear on multiple physical pins until the default configuration).

### DIFF
--- a/rtl/system/pinmux.sv
+++ b/rtl/system/pinmux.sv
@@ -115,7 +115,7 @@ module pinmux
 
   always @(posedge clk_i or negedge rst_ni) begin
     if (!rst_ni) begin
-      // Select second input by default so that pins are connected to the first block that is specified in the configuration.
+      // Select second mux input by default so that pin is connected to the first block that is specified in the configuration.
       ser0_tx_sel <= 2'b10;
     end else begin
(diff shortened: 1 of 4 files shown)


GregAC:
I'll rebase this on the pinmux PR when it appears

marnovandermaas (commented):
From a first glance this looks reasonable. I think I would prefer to have this in the RS-485 PR so that I can see what the effects are in the RTL when the no default option is used.

GregAC:
Ok I'll refactor the RS485 PR so the RTL change can be clearly seen

GregAC:
@marnovandermaas https://github.com/lowRISC/sonata-system/pull/313 has been refactored to show the RTL change using this feature, relevant commit is https://github.com/lowRISC/sonata-system/pull/313/commits/3453c8be4f3dc0f2d72404463764c63622dba814

marnovandermaas:
This has been included in the RS485 PR.

PR #307, opened 2024-10-31
Add PWM tests

AlexJones0:
This PR implements a few PWM tests for Sonata which can be run in both FPGA and in simulation, and should be able to run in CI with an appropriate jumper cable connection. See the commit messages for more details.

Requires a jumper cable connecting the MikroBus PWM and PMOD0 pin 1 on the board to pass. One test sets the period and duty cycles to a range of different values and measures the PWM output via the GPIO to check that these values are being correctly output, given a certain margin of error due to the limited size of the counter. In practice it is found that the test consistently passes for hundreds of iterations when given around 20 cycles margin on each side, which is granular enough that we can still be confident that the PWM is functioning as intended by setting a variety of periods and duty cycles.

The other test simply checks that when the period/counter is configured to 0, the PWM always outputs 0 (low).

This has been tested to pass both on FPGA and in simulation by running it using the normal procedures, and has been tested to pass for 100 iterations on FPGA (and with a variety of different parameters in manual testing) when run locally.


sw/cheri/tests/pwm_tests.hh
```diff
@@ -0,0 +1,207 @@
...
+ * @param period The PWM period to test with
+ * @param duty_cycle The PWM duty cycle to test with. Represented as a separate counter, so for example a duty cycle of
+ * 50 with a period of 200 would give 1/4 i.e. 25% high output.
+ * @param cycles The number of PWM cycles to test
+ * @param error_delta The margin of error on either side of the number of cycles measured for the period and duty cycle.
+ * There will always be some error, as measuring logic cannot run on every cycle.
+ *
+ * @returns The integer number of failures during the test
+ */
+int pwm_loopback_test(Capability<volatile SonataGpioPmod> gpio_pmod0, Capability<volatile SonataPwm> pwm,
```
engdoreis, line 59: Isn't there aliases for these capabilities in `common/sonata-devices.hh`

engdoreis (approved):

sw/cheri/tests/pwm_tests.hh
AlexJones0: No, `SonataGpioPmod` is a part of the new GPIO driver and I don't want to make sweeping changes to use that here because it will likely soon be replaced by `SonataGpioPmod0` after the new pin mapping changes are merged. Likewise, `SonataPwm` hasn't been used before so I suppose I could add an instance of it to `sonata-devices.hh`, but again I think that is better addressed in a follow-up PR which cleans up some of these instances, because the new mapping changes will change the PWM driver to use 7 instances instead of just 1.

PR #311, opened 2024-11-01
Rework Vivado optimisations for wide compatibility

elliotb-lowrisc:
Turns out that Vivado 2024.1 rearranges its implementation script so that reporting is done before the post-route hook is called. This made it seem as though builds were failing timing, even when the post-route hook optimisations had fixed them.

Exploration revealed that Vivado has an optional post-route optimisation stage that can be enabled and tweaked via TCL scripts. This includes updated reporting and such.
So, rework the post-route optimisation to use Vivado's stage rather than a custom hook.

This required putting the optimisation level in vivado_setup.tcl rather than in one of the hooks, which meant that all the optimisation level decision making also had to move to the same.

This should be effectively a no-op except for reporting.

Not to be merged before 1.0 release.


elliotb-lowrisc:
Rebase and add update of CI timing report path

marnovandermaas (commented):
I can confirm that this passes timing on 2024.1, will wait for CI to finish running on top of the rebase before approving.

marnovandermaas (approved):
CI seems passing too so I'm merging this.

PR #312, opened 2024-11-01
Add PWM Always High test & Misc. Improvements

AlexJones0:
Addresses feedback from @alees24 after #307 was merged.

Makes some small improvements to the PWM tests. First adds an "Always high" PWM test, which checks that the PWM always outputs 1 when the duty cycle is in some configuration greater where it is greater than the period/counter. This acts as an opposite to the "Always Zero" test.

Second, this PR adds a couple more PWM loopback tests cases, to better cover the range of duty cycles above 50%, which were previously not being tested. These are 224/255 and 192/255.

Third, corrects a couple of typos.

This has been tested to pass on FPGA and CI locally.

### DIFF
--- a/sw/cheri/tests/pwm_tests.hh
+++ b/sw/cheri/tests/pwm_tests.hh
@@ -17,7 +17,7 @@ typedef SonataPulseWidthModulation SonataPwm;
 
 /**
  * Configures the number of test iterations to perform.
- * This can be overriden via a compilation flag.
+ * This can be overridden via a compilation flag.
  */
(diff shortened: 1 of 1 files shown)


alees24 (approved):
LGTM, thank you.

AlexJones0:
Force push should have fixed a small bug I missed where on the last tested period (`254`), the duty cycle calculated would have been equal to and not greater than the period, meaning that there would in theory be a 1/255 chance of this test failing (albeit perhaps this may not be the case due to the consistency of timing between runs from setting the PWM to measuring it).

PR #309, opened 2024-11-01
New Mappings

HU90m:

PR #313, opened 2024-11-03
Rs485

GregAC:


GregAC:
Sits on top of https://github.com/lowRISC/sonata-system/pull/306 and includes those commits, https://github.com/lowRISC/sonata-system/pull/313/commits/3e0434796d8c07e842673a15e801c39d582f0e27 is the only relevant one for reviewing.

Will need to be rebased on https://github.com/lowRISC/sonata-system/pull/309 when that's merged with the UART used adjusted at the same time (move from UART 4 to UART 2).

elliotb-lowrisc:
XDC changes look fine to me. Nice idea to constraint the number of cycles to the smallest number required by calculation.

GregAC:
Rebased on pinmux changes, now connected to UART 2.

Just realised I haven't updated the pinmux dotfile etc but can always do a follow-up for it

data/pins_sonata.xdc
```diff
@@ -140,6 +140,12 @@ set_property -dict { PACKAGE_PIN G18 IOSTANDARD LVCMOS33 } [get_ports ser1_rx]
 set_property -dict { PACKAGE_PIN P3  IOSTANDARD LVCMOS33 } [get_ports rs232_tx]
 set_property -dict { PACKAGE_PIN N6  IOSTANDARD LVCMOS33 } [get_ports rs232_rx]
 
+## UART RS485
```
marnovandermaas, line 143: Nit: RS-485

data/synth_timing.xdc
```diff
@@ -400,6 +400,30 @@ set_input_delay -clock vclk_sys -max 0 [get_ports rs232_rx]
 set_input_delay -clock vclk_sys -min 0 [get_ports rs232_rx]
 set_output_delay -clock vclk_sys -max 0 [get_ports rs232_tx]
 set_output_delay -clock vclk_sys -min 0 [get_ports rs232_tx]
+## UART RS232
```
marnovandermaas, line 403: Nit: RS-232

rtl/fpga/top_sonata.sv
```diff
@@ -492,4 +509,37 @@ module top_sonata
       scl0
     })
   );
+
+  // 90ns switch time + 5ns FPGA delay output time. If this parameter is adjusted constraints on
+  // rs485_de/rs485_ren in synth_timing.xdc must be adjusted to match
+  parameter int unsigned Rs485TransceiverSwitchDelayNs = 90 + 5;
+
+  // Calculate system clock cycles required to cover Rs485TransceiverSwitchDelayNs. Uncondtionally
```
marnovandermaas, line 517: Nit: Unconditionally

rtl/system/rs485_ctrl.sv
```diff
@@ -0,0 +1,187 @@
...
+  output logic di_o,
+  output logic ren_o,
+  output logic de_o,
+  input  logic ro_i
+);
+  // Time to wait in cycles before disabling the driver after transmission finished. Needs to factor
+  // in TransceiverSwitchCycles as the transmission is delayed by that many cycles through the
+  // `di_delay` shift register.
+  localparam int unsigned TransmitCooldownCount = TransceiverSwitchCycles + TransmitEndCycles;
+  localparam int unsigned CounterW = $clog2(TransmitCooldownCount);
```
marnovandermaas, line 51: Do we need to make this `$clog2(TransmitCooldownCount + 1)`?

rtl/system/rs485_ctrl.sv
```diff
@@ -0,0 +1,187 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// Controls the RS485 transceiver. RS485 is half-duplex so cannot transmit and receive
```
marnovandermaas, line 5: Nit: RS-485

rtl/system/sonata_system.sv
```diff
@@ -1335,6 +1338,21 @@ module sonata_system
...
+
+  // Enable the RS485 receiver any time the RS485 RX input is connect to the UART via pinmux.
+  // Whether or not we are transmitting does not need to be factored in here as that is dealt with
+  // via rs485_ctrl.
+  assign rs485_rx_enable_o = u_pinmux.uart_rx_2_sel[4];
+
+  // Transmission enabled when UART is muxed to RS485 TX output and UART is actively transmitting.
+  assign rs485_tx_enable_o = u_pinmux.rs485_tx_sel[1]                           &
+                             gen_uart_blocks[2].u_uart.uart_core.tx_enable      &
+                             ~gen_uart_blocks[2].u_uart.hw2reg.status.txidle.d;
```
marnovandermaas, line 1354: We should open an issue that makes sure this does not have to be manually updated if the mapping of RS-485 in the top config changes.

GregAC:
Timing looking good on this when rebased on main

sonata.core
```diff
@@ -30,6 +31,7 @@ filesets:
       - lowrisc:dv_dpi_c:usbdpi:0.1
       - lowrisc:dv_dpi_sv:usbdpi:0.1
     files:
+      - rtl/system/rs485_ctrl.sv: { file_type: systemVerilogSource }
```
marnovandermaas, line 34: Why do you need to add it here again when it's already in files_sonata?

marnovandermaas (commented):
Left a few comments on this PR.

rtl/system/rs485_ctrl.sv
GregAC: Yes we do, in fact leaving this off led to incorrect behaviour which I'd identified and thought I'd fixed, must have accidentially killed the change in a rebase, well spotted!

rtl/system/sonata_system.sv
GregAC: I was thinking we could add a feature to pinmux where you could export certain select signals out of the pinmux so we can both ditch some of the hierarchical references and the specific pinmux signal names here.

Perhaps a similar feature that allows you to generate specific parameters that tell you which block (index wise) a particular pin can be muxed to as well so we can drop the `2` here and replace with `RS485_UART` that gets updated with the pinmux generation.

sonata.core
GregAC: `files_verilator` is used for the verilator build and `files_sonata` is only used for the FPGA build.

`rs485_ctrl` is a little awkward as really it'd sit in `sonata_system` but because of the way we've got the pinmux setup (and it needs to sit on top of certain pins from pinmux) it's neater to integrate into `top_sonata` and so it's separately included in top_verilator`

GregAC:
Thanks for the review @marnovandermaas I believe I've addressed your comments. Also adjusted timing constraints to make timing easier on the RS485 transceiver IO

I running a local synthesis now with exactly the commit for this PR and will rerun the RS485 tests one last time (already checked after I fixed up the `TransmitCooldownCount` issues but best double check given I messed that up last time!).

rtl/system/sonata_system.sv
alees24: I agree that it would definitely be preferable to tidy this up, and given that the uart presently does this...
```
  // always enable the driving out of TX
  assign cio_tx_en_o = 1'b1;
```
... it would seem sensible to drive out an actual 'tx_en_o' signal for use here.

GregAC:
Ok tests all looking good and with the timing constraint changes we have 22ns slack (positive) on the `rs485_re` and `rs485_den` outputs.

GregAC:
@marnovandermaas happy for this to go in?

alees24 (approved):
I've tested transmit and receive via the uartdpi in Verilator sim (from a direct fetch of the PR) using `rs485_check` and transmit on the FPGA using a logic analyser with improvised pull ups/termination resistors.
Additionally I've inspected the RTL logic in some detail.

It would be good at some point to address the cross-module references, external (outside the uart) generation of 'uart_tx_en', and the duplication of rs485_ctrl in both top_verilator and top_sonata, please.

GregAC:
Great thanks for your review @alees24

sw/cheri/checks/rs485_check.cc
```diff
@@ -0,0 +1,77 @@
...
+    char uart_capture_buf[256];
+    uint32_t chars_seen = 0;
+    char ch;
+
+    do {
+      ch = rs485_uart->blocking_read();
+
+      write_str(uart, "Saw RS-485: 0x");
+      write_hex8b(uart, ch);
+      write_str(uart, "\r\n");
```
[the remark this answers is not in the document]
GregAC: Agreed, this will fall over quickly if there's transmission happening at any speed, I've just used it for interactive use as a basic receive/transmit test (one that proved useful as I determined I had to let the `tx` line return to idle state before disabling the driver otherwise the transceiver doesn't correctly output idle from the receiver when the RS485 is un-driven).

I was considering writing a higher performance test that used two sonata board along with a PRNG to generate a stream of predictable numbers that can be sent from one board and received by another and checked.

PR #314, opened 2024-11-03
Perf improvements

GregAC:
Proposed performance improvements (improving SRAM throughput and Ibex single cycle multiplier), rebased on top of the pinmux changes (https://github.com/lowRISC/sonata-system/pull/309).

Timing is passing and build time looking good.

Two things were required to achieve this:

1. Increasing the optimization level (https://github.com/lowRISC/sonata-system/pull/302), though this had no clear impact on build times
2. Adding a register stage to the internal SPI loopback (see commit https://github.com/lowRISC/sonata-system/commit/dd850880282427653548ed9738fa62a015198450). The timing path this cuts is odd, there's a single level of logic between two registers with almost 25ns worth of delay. I suspect it's something to do with the SPI timing constraints causing the tool to want the register to output at a hugely skewed time. We may just be on the edge of getting away with it and these changes have pushed it over the edge so regardless of whether we take these changes we should likely cut the internal loopback path.

I've also added saving the utilization and timing reports from the FPGA CI as artifacts so they can be examined after (in particular so you can look at the timing results for any particular PR).


GregAC:
Note that you need an RS485 -> USB convertor to run the check programs on the FPGA but they can also be run on verilator (I've added a new DPI UART you can connect to for the RS485).

The `rs485_check` will listen for incoming messages on the rs485 and echo them back, `rs485_spam_check` just sends a changing message out over and over again on the rs485.

It's possible a second sonata board could be used for testing (e.g. running a program that just echos bytes received on rs485 back on the system UART) but I haven't had time to explore this.

GregAC:
For reference here is the timing summary  from the CI run (from the implementation reports artefact)

```
    WNS(ns)      TNS(ns)  TNS Failing Endpoints  TNS Total Endpoints      WHS(ns)      THS(ns)  THS Failing Endpoints  THS Total Endpoints     WPWS(ns)     TPWS(ns)  TPWS Failing Endpoints  TPWS Total Endpoints  
    -------      -------  ---------------------  -------------------      -------      -------  ---------------------  -------------------     --------     --------  ----------------------  --------------------  
      0.043        0.000                      0                42250        0.066        0.000                      0                42249        1.741        0.000                       0                 15674  
```

Here's the utilization report

```
+----------------------------+-------+-------+------------+-----------+-------+
|          Site Type         |  Used | Fixed | Prohibited | Available | Util% |
+----------------------------+-------+-------+------------+-----------+-------+
| Slice LUTs                 | 22736 |     0 |          0 |     32600 | 69.74 |
|   LUT as Logic             | 22628 |     0 |          0 |     32600 | 69.41 |
|   LUT as Memory            |   108 |     0 |          0 |      9600 |  1.13 |
|     LUT as Distributed RAM |   108 |     0 |            |           |       |
|     LUT as Shift Register  |     0 |     0 |            |           |       |
| Slice Registers            | 15330 |     0 |          0 |     65200 | 23.51 |
|   Register as Flip Flop    | 15330 |     0 |          0 |     65200 | 23.51 |
|   Register as Latch        |     0 |     0 |          0 |     65200 |  0.00 |
| F7 Muxes                   |   551 |     0 |          0 |     16300 |  3.38 |
| F8 Muxes                   |   142 |     0 |          0 |      8150 |  1.74 |
+----------------------------+-------+-------+------------+-----------+-------+
```

Build time was 11m 30s which is in line with previous build times in CI

elliotb-lowrisc:
I've also seen an LCD SPI path with 24 ns of delay on a single net. The hold slack on the path was only 13 ns, so I suppose it might be hold-fixing gone wrong.

.github/workflows/ci.yml
```diff
@@ -86,6 +86,8 @@ jobs:
     runs-on: [ubuntu-22.04-fpga, sonata]
     env:
       BITSTREAM_PATH: build/lowrisc_sonata_system_0/synth-vivado/lowrisc_sonata_system_0.bit
+      TIMING_RPT: build/lowrisc_sonata_system_0/synth-vivado/lowrisc_sonata_system_0.runs/impl_1/top_sonata_timing_summary_routed.rpt
```
elliotb-lowrisc, line 89: Is there a way to make this conditional on whether "top_sonata_timing_summary_postroute_physopted.rpt" exists, so we avoid looking at the wrong report when we switch to the new 2024.1 post-route optimisation flow?

rtl/ip/spi/rtl/spi.sv
```diff
@@ -187,7 +187,13 @@ module spi import spi_reg_pkg::*; #(
   // Internal loopback functionality allowing the input (CIPO) to be received directly from
   // the output (COPI) for testing.
   logic spi_cipo;
-  assign spi_cipo = reg2hw.control.int_loopback.q ? spi_copi_o : spi_cipo_i;
+  logic spi_copi_q;
+
+  always_ff @(posedge clk_i) begin
```
elliotb-lowrisc, line 192: I have noticed the lack of reset but think it is fine. The `u_spi_core/copi_shift_q` driving it also lacks a reset, so this is just replicating the behaviour one cycle later.

GregAC: yeah I prefer to leave resets off flops that don't strictly need them. Given the SPI ignores incoming data when it's not been specifically command to clock things in/out then it won't ever actually 'see' the first bit out of reset anyway.

elliotb-lowrisc:
I've run a bitstream build of this PR and a baseline build with the changes it is based on (up to and including the opt=2 change) using Vivado 2021.1 on my local machine. The breaking of the SPI loopback path did seem to improve the overall WNS (-1.451 -> -0.391). Though, for these two builds at least, the TNS went the other way (-8.168 -> -27.834), though the numbers are still relatively small. Build times were also increase substantially (19m -> 32m) though that is muddied by builds going on in parallel. All this is somewhat concerning (for the state of Sonata if not this PR) but not conclusive. More testing to follow

elliotb-lowrisc:
Adding timing constraint updates (https://github.com/lowRISC/sonata-system/pull/316) and building ono showed a more reasonable build time (15m) and timing passing.

GregAC:
Thanks for checking it out @elliotb-lowrisc. Frustrating it's giving you some failures.

I've certainly see some flakey behaviour from the implementation flow, one version of this (pre pinmux changes) actually gave a noticeable improvement in timing, taking us to 0.5 ns WNS (positive)!

I've rebased this locally now the pinmux changes are in and am doing another run, will see what results!

elliotb-lowrisc:
I think the SPI loopback change is definitely worth a quick PR and merge if possible

GregAC:
> I think the SPI loopback change is definitely worth a quick PR and merge if possible

Agreed I'll spin out a separate PR

PR #315, opened 2024-11-04
Revert device alignment reductions

marnovandermaas:
This reverts commit 90540001ccd6b8d38b2b85d8a69a68c755c6770d.

This reversion avoids increasing the pressure of decoding addresses on the crossbar.


GregAC (approved):
Looks reasonable to me, was this motivated by a particular issue/analysis or just a desire to generally ease timing in this area?

HU90m (changes requested):
We should probably revert both this and `a75d83616fd22d63860ba1bd9cca17f2fb692d69` in one, no?

marnovandermaas:
> We should probably revert both this and `a75d83616fd22d63860ba1bd9cca17f2fb692d69` in one, no?

Good point, I had forgotten it was lowered in two steps. This should be in now.

PR #318, opened 2024-11-04
Pmod rename

marnovandermaas:
This renames the pmod pins to be conformant with the digilent specification.


HU90m:
Have you chatted to @AlexJones0 about this?

marnovandermaas:
I briefly told him this in person and it seemed like he's on board. He hasn't seen the code though.

HU90m:
Alex had suggested this before. I was originally trying to stick to the names on the sonata board schematic and I was apprehensive about having another set of pin names, but I think you all are right. We should breaking from the schematic names for a nicer user experience.

While I was changing the top config in https://github.com/lowRISC/sonata-system/pull/326, I've removed the inconsistent spi suffixes from the Pi Hat pins and changed to the standard physical pin numbering for Pmod.

marnovandermaas:
Closed as it is replaced in #326

PR #319, opened 2024-11-04
[ci] Save bitstream utilization and timing reports as artifacts

GregAC:


GregAC:
As @elliotb-lowrisc has pointed out this will require a change when we switch  to v2024.1 as log files get renamed but I think it's worth chucking in now, will support final v1.0 work especially as timing is becoming critical.

GregAC:
Note there's no artifacts produced by CI there's no bitstream build, you can see the same change successfully producing the correct artifacts here: https://github.com/lowRISC/sonata-system/actions/runs/11655196636

marnovandermaas (commented):
Any idea why we're getting this warning on this PR:
`Warning: No files were found with the provided path: build/lowrisc_sonata_system_0/synth-vivado/lowrisc_sonata_system_0.runs/impl_1/top_sonata_timing_summary_routed.rpt`

GregAC:
> Any idea why we're getting this warning on this PR: `Warning: No files were found with the provided path: build/lowrisc_sonata_system_0/synth-vivado/lowrisc_sonata_system_0.runs/impl_1/top_sonata_timing_summary_routed.rpt`

Yes, that's what my comment above referred to, CI doesn't build a bitstream so it warns it can't find the artifact path. Happily (in this case) this doesn't result in a CI failure so we don't need to worry about handling this case.

HU90m (commented):
Does this correctly handle the case where a bitstream isn't built (e.g. is the upload skipped)?

HU90m (approved):

PR #320, opened 2024-11-04
[rtl] Add register stage to SPI internal loopback

GregAC:
Timing failures have been observed on this loopback path, it has minimal logic levels but a very long delay, possibly due to I/O timing constraints.

Adding the register stage cuts the internal path. With the register stage the internal loopback cannot run at full speed, however as this is for testing purposes only this is acceptable.

Bizarrely causes timing failures to introduce this right now, so not to be merged yet until we get on top of the timing instability.

### DIFF
--- a/rtl/ip/spi/rtl/spi.sv
+++ b/rtl/ip/spi/rtl/spi.sv
@@ -187,7 +187,13 @@ module spi import spi_reg_pkg::*; #(
   // Internal loopback functionality allowing the input (CIPO) to be received directly from
   // the output (COPI) for testing.
   logic spi_cipo;
-  assign spi_cipo = reg2hw.control.int_loopback.q ? spi_copi_o : spi_cipo_i;
+  logic spi_copi_q;
+
+  always_ff @(posedge clk_i) begin
+    spi_copi_q <= spi_copi_o;
+  end
+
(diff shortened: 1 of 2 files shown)


alees24:
I had the same/similar failing here just now, hence the discovery of the untethered `cipo_i` (PR #321) signal on that LCD-specific SPI controller; now, one would expect that it just gets tied low as an implicit net but it's also hard to ignore the fact that it's that one specific signal with the weird 25ns delay.

elliotb-lowrisc:
I still have no good explanation for why this odd path delay happens. Only the observations that I've only ever seen it on the LCD SPI, that other loads on the net may have much shorter and more reasonable paths (don't use the long winding path segment), and that it has tight hold timing due its nature as a shift register with minimal logic on the shift path. My best guess is that it was something to do with how the tool handled the unconnected input, but I have no solid basis beyond coincidence and lack of faith in the tool.

GregAC:
Worth investigating this properly (maybe with the unconnected input sorted this has gone away?) but not something we need for v1.0 now as it's not causing a critical timing issue

elliotb-lowrisc:
See https://github.com/lowRISC/sonata-system/issues/330

GregAC:
Fixed with https://github.com/lowRISC/sonata-system/pull/334 so register stage won't be required

PR #321, opened 2024-11-05
Explicitly tie off unused inputs

alees24:


marnovandermaas (approved):
This looks correct to me.

PR #322, opened 2024-11-05
Legacy software fixup

marnovandermaas:
This fixes up the hello world demo application and at least partially fixes the ethernet test.


sw/legacy/common/rv_plic.c
```diff
@@ -7,7 +7,9 @@
 #include "dev_access.h"
 #include "sonata_system.h"
 
-#define NUM_IRQS 182
+#define NUM_IRQS 32
+
+#define I2C_ACQ_THRESHOLD_INTERRUPT 16
```
AlexJones0, line 12: Shouldn't this be called something like `I2C0_INTERRUPT` instead of `I2C_ACQ_THRESHOLD_INTERRUPT`, because we only have a single interrupt line per block now?

sw/legacy/demo/hello_world/main.c
```diff
@@ -99,19 +99,13 @@ int main(void) {
...
-        out_val          = (out_val << 1) & GPIO_LED_MASK;
-        if ((in_val & 0x1) || (out_val == 0)) {
-          out_val = 0x10;
-        }
-        set_outputs(GPIO_OUT, out_val);
+      // Cycle through LEDs
+      uint32_t out_val = read_gpio(GPIO_OUT);
+      out_val          = (out_val << 1) & GPIO_LED_MASK;
+      if (out_val == 0) {
+        out_val = 0x1;
```
AlexJones0, line 106: Why the functionality change here? Would it not be better to change it to 
```c
if ((in_val & (1u << 9)) || (out_val == 0)) {
```
to retain the ability to reset the LEDs by pressing up on the joystick?

AlexJones0: It probably was (I am unaware of the history of this demo), but in theory by reading GPIO input 1 it would have previously also worked when moving the joystick up, as the joytick navigation switches used to be on GPIO0-4 before the recent pin mapping changes. If you don't think this behaviour is relevant any more then feel free to leave it as is

sw/legacy/common/gpio.h
```diff
@@ -12,8 +12,8 @@
 #define GPIO_IN_DBNC_REG 0x8
 #define GPIO_OUT_SHIFT_REG 0xC
 
-#define GPIO_OUT_MASK 0x7FFF  // Support 15-bit output
-#define GPIO_LED_MASK 0x0FF0  // These 8 bits are the user LEDs
+#define GPIO_OUT_MASK 0x01FF  // Support 15-bit output
```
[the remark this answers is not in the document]
marnovandermaas: Thanks, I've actually removed this define because it isn't used anywhere.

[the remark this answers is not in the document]
marnovandermaas: But you're right my mask was wrong. It should have been equal to the LED one.

sw/legacy/demo/hello_world/main.c
```diff
@@ -99,19 +99,13 @@ int main(void) {
       // Re-enable interrupts with output complete
       arch_local_irq_enable();
 
-      // Cycling through green LEDs
-      if (USE_GPIO_SHIFT_REG) {
```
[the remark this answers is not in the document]
marnovandermaas: Yes!

sw/legacy/demo/hello_world/main.c
marnovandermaas: Yes, I've added this in. Thanks for the suggestion!

sw/legacy/common/rv_plic.c
marnovandermaas: Yes, have renamed this now.

AlexJones0 (approved):
LGTM now, thanks @marnovandermaas.

PR #323, opened 2024-11-05
Resurrect SD card operation

alees24:
Combining CIPO lines directly without regard to Chip Selects is risky, and in particular the line `appspi_d1` from the flash is tri-stated without a pullup resistor on the FPGA board.

Modifications to the top_verilator implmentation to match the top_sonata implementation more closely, and paving the way for the introduction of the microSD card DPI model (already built and operational).

_Without these changes the flash works only by virtue of the fact the the Sonata FPGA board carries a pullup on microSD card CIPO line, and the SD card is unusable because the flash CIPO does not. With these changes in place, and appropriate updates to the software that drives the microSD card and LCD, both the reading of SD card filing systems and the basic video player are operational once more._

@elliotb-lowrisc FPGA build remains timing clean (v2022.2), but it would be worth giving it the once over, please.

### DIFF
--- a/data/pins_sonata.xdc
+++ b/data/pins_sonata.xdc
@@ -291,6 +291,7 @@ set_property -dict { PACKAGE_PIN C10 IOSTANDARD LVCMOS33 } [get_ports appspi_d1]
 set_property -dict { PACKAGE_PIN A10 IOSTANDARD LVCMOS33 } [get_ports appspi_d2]
 set_property -dict { PACKAGE_PIN A9  IOSTANDARD LVCMOS33 } [get_ports appspi_d3]
 set_property -dict { PACKAGE_PIN D10 IOSTANDARD LVCMOS33 } [get_ports appspi_cs]
+set_property PULLTYPE PULLUP [get_ports appspi_d1]
 
 ## Ethernet MAC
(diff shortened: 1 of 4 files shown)


GregAC (changes requested):
Thanks for spotting this @alees24 but I think I'd prefer we use the pinmux for this.

I can see the logic in doing this shared bus arrangement but two reasons to use the pinmux 

1. We end up having to implement a custom pinmux like arrangement for the incoming data lines
2. Anything with access to the SPI instance will be able to access both flash and SD card. If we pinmux things you can come up with an arrangement where a compartment gets temporary access to the SPI capability but not access to the pinmux itself so can ensure it only gets access to flash or the SD card but not both.

alees24:
I'm fine with CIPO selection being returned to the pinmux; it's only implemented in sonata_system presently because _all_ of the signals to/from that SPI controller have been removed from the pinmux.

GregAC:
On point 2. I imagining most uses of the pinmux will see things get setup immediately after RTOS boot so for this SPI you'd mux to the SD card and leave the flash non-connected. This then gives the  pleasing property that the software cannot modify the boot medium.

elliotb-lowrisc:
For what it's worth, the current changes don't seem to have caused a drop in overall QoR when building with Vivado 2021.1

marnovandermaas:
Closing this as it's been solved by #326

PR #324, opened 2024-11-05
Fix SPI controller interrupt routing

alees24:
With the semantic change to the `SPI_NUM` parameter, the final 3 SPI controllers lost their ability to stimulate the PLIC. Correct the wiring.

(Built into FPGA on top of PR #323 and timing-clean; tests/test_runner passing.)


HU90m (approved):
@engdoreis I'm concerned that this wasn't caught by the interrupt test. I've created https://github.com/lowRISC/sonata-system/issues/325

marnovandermaas (commented):
I would have probably preferred to define another local parameter equal to `SPI_NUM + 3` instead of hardcoding the 3 in multiple places.

PR #327, opened 2024-11-06
Remove unused variables

elliotb-lowrisc:


AlexJones0 (approved):
LGTM when CI passes, thanks @elliotb-lowrisc.

PR #328, opened 2024-11-06
Make PLIC test fail when interrupt not fired

AlexJones0:
This PR makes sure that the `irq_fired` attribute is correctly reset to `false` after each `wfi()` call in a test, which was previously causing the test to pass even when the interrupt lines for a block were not correctly hooked up in hardware.

As that would cause the entire test runner to halt, this commit then replaces the ASM `wfi` functionality with a basic timeout for handling a fired interrupt, which is set to 10 microseconds by default so as to not cause lengthy waits when running in simulation. This timeout means that the `wfi` function can actually terminate without an interrupt being fired, and that the tests can now query the `irq_fired` attribute to determine whether the test passed or failed, and log accordingly.

When tested locally against a bitstream built on d962657f501449d6e5517d21c3e08e5d8b8e95c4, this test correctly fails as expected with the following output: 

```
  Testing SPI0:
  irq fired: 0x18, expected: 0x18 : IP-local fired: 0x10, expected: 0x10... PASS!
  irq fired: 0x18, expected: 0x18 : IP-local fired: 0x1, expected: 0x1... PASS!
  irq fired: 0x18, expected: 0x18 : IP-local fired: 0x2, expected: 0x2... PASS!
  irq fired: 0x18, expected: 0x18 : IP-local fired: 0x4, expected: 0x4... PASS!
  irq fired: 0x18, expected: 0x18 : IP-local fired: 0x8, expected: 0x8... PASS!
  Testing SPI1:
  irq fired: 0x19, expected: 0x19 : IP-local fired: 0x10, expected: 0x10... PASS!
  irq fired: 0x19, expected: 0x19 : IP-local fired: 0x1, expected: 0x1... PASS!
  irq fired: 0x19, expected: 0x19 : IP-local fired: 0x2, expected: 0x2... PASS!
  irq fired: 0x19, expected: 0x19 : IP-local fired: 0x4, expected: 0x4... PASS!
  irq fired: 0x19, expected: 0x19 : IP-local fired: 0x8, expected: 0x8... PASS!
  Testing SPI2:
  no irq fired within timeout for irq: 0x1a : IP-local: 0x10... FAIL!
  no irq fired within timeout for irq: 0x1a : IP-local: 0x1... FAIL!
  no irq fired within timeout for irq: 0x1a : IP-local: 0x2... FAIL!
  no irq fired within timeout for irq: 0x1a : IP-local: 0x4... FAIL!
  no irq fired within timeout for irq: 0x1a : IP-local: 0x8... FAIL!
  Testing SPI3:
  no irq fired within timeout for irq: 0x1b : IP-local: 0x10... FAIL!
  no irq fired within timeout for irq: 0x1b : IP-local: 0x1... FAIL!
  no irq fired within timeout for irq: 0x1b : IP-local: 0x2... FAIL!
  no irq fired within timeout for irq: 0x1b : IP-local: 0x4... FAIL!
  no irq fired within timeout for irq: 0x1b : IP-local: 0x8... FAIL!
  Testing SPI4:
  no irq fired within timeout for irq: 0x1c : IP-local: 0x10... FAIL!
  no irq fired within timeout for irq: 0x1c : IP-local: 0x1... FAIL!
  no irq fired within timeout for irq: 0x1c : IP-local: 0x2... FAIL!
  no irq fired within timeout for irq: 0x1c : IP-local: 0x4... FAIL!
  no irq fired within timeout for irq: 0x1c : IP-local: 0x8... FAIL!
  Testing USBDEV:
  ...
```

Since #324 this has now been fixed on `main` and as such this test should pass in CI.

### DIFF
--- a/sw/cheri/tests/plic_tests.hh
+++ b/sw/cheri/tests/plic_tests.hh
@@ -12,6 +12,7 @@
 #include "../../common/defs.h"
 #include "../common/asm.hh"
 #include "../common/sonata-devices.hh"
+#include "../common/timer-utils.hh"
 #include "test_runner.hh"
 
 #include <array>
@@ -42,6 +43,11 @@ struct PlicTest {
 
   bool is_irq_clearable;
   volatile bool irq_fired = false;
+
+  // Rounded CPU clock cycles per microsecond, as this may not be an integral number
+  static constexpr uint32_t CyclesPerUsec = (CPU_TIMER_HZ + 999'999) / 1'000'000;
+  // The number of cycles to wait for an interrupt before timing out.
+  static constexpr uint32_t wfiTimeout = 10 * CyclesPerUsec;
 
   void (*irq_handler)(PlicTest *plic_test, PLIC::Interrupts irq) = nullptr;
 
@@ -99,6 +105,11 @@ struct PlicTest {
       uart->interrupt_enable(uartMap[i].id);
       uart->interruptTest = ip_irq_id;
       wfi();
+      if (!irq_fired) {
+        error_count++;
+        log(static_cast<PLIC::Interrupts>(0), 0, true);
+      }
+      irq_fired = false;
     }
   }
 
@@ -163,6 +174,11 @@ struct PlicTest {
       i2c->interrupt_enable(i2cMap[i].id);
       i2c->interruptTest = 0x01 << ip_irq_id;
       wfi();
+      if (!irq_fired) {
+        error_count++;
+        log(static_cast<PLIC::Interrupts>(0), 0, true);
+      }
+      irq_fired = false;
     }
   }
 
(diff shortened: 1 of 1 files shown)


sw/cheri/tests/plic_tests.hh
```diff
@@ -99,6 +105,11 @@ struct PlicTest {
       uart->interrupt_enable(uartMap[i].id);
       uart->interruptTest = ip_irq_id;
       wfi();
+      if (!irq_fired) {
+        error_count++;
+        log(static_cast<PLIC::Interrupts>(0), 0, true);
+      }
+      irq_fired = false;
```
alees24, line 112: This clean up code is common to all of the IP blocks; I think it'd be safer to move it into the wfi() function itself, please, so that it remains present and correct for all IP blocks. Also, may I suggest that wfi() could perhaps be renamed so that it doesn't look quite so like a compiler intrinsic/assembler instruction; the casual reader may not spot the member function.

PR #329, opened 2024-11-06
Clear IRQ in the event of a failure

alees24:
Extend the 'IRQ did not fire' improvement to tidy up the interrupt state in the event of a failure to avoid confusion with follow-on problems.

If the `intr_test` or `intr_enable` register(s) remain non-zero then subsequent test runs without an intervening system reset will behave unexpectedly.


marnovandermaas (commented):
The code changes look reasonable to me, but need to wait for CI to pass and Alex to have a look.

AlexJones0 (approved):
Thank you for adding this @alees24, it all looks good to me.

PR #331, opened 2024-11-06
Updated to latest RTOS fork and lowrisc-nix

HU90m:
- Updated to latest version of lowRISC RTOS fork
- Updated lowrisc-nix version
  - We'll need the latest version of the compiler to be able to build the
upstream CHERIoT RTOS.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -395,11 +395,11 @@
         "rust-overlay": "rust-overlay_3"
       },
       "locked": {
-        "lastModified": 1727339002,
-        "narHash": "sha256-h/daSpZNRs+HkUMAEQwqTs/GHrcIYo+ZPUEaghuAfeY=",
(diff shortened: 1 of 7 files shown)


AlexJones0:
Force pushes have rebased on latest `main`, and updated the software (usbdev utils, usbdev test, plic test) to use the refactored USB device driver so that all tests should still work.

HU90m (commented):
My own PR so I can't give it a green tick, but I would if I could. Thanks Alex. Do merge

AlexJones0 (approved):
Approving because I'm happy with @HU90m's build changes and he appears to be happy with my corresponding software changes.

marnovandermaas (approved):
Looks reasonable to me. We still need some more updates in our RTOS fork, but we can always bump it later.

PR #332, opened 2024-11-06
Ensure mikroBUS Click RST/RE pin is driven

alees24:
This RST pin is an active low reset signal to most mikroBUS Click boards, but some boards use it as an active low Read Enable (RE) pin, so presently not all boards are supported.

Avoid producing undefined behaviour on (hopefully) the majority of boards, and rely upon power-cycling and software-reset mechanisms.


HU90m (approved):
This'll do nicely for now. Thanks Adrian!

PR #334, opened 2024-11-07
Insert delay on lcd_copi to avoid timing issue

elliotb-lowrisc:
Manually insert delay in the form of BUF (LUT1) cells directly before the lcd_copi output buffer (OBUF) to satisfy the large hold requirement of the LCD driver chip.

If left to add hold-fixing itself, the tool occasionally causes setup timing failures on internal shift/loopback paths by placing the hold-fixing (in the form of a hold detour) too early in the path. Hold fixing as large as this should be put close to the port, after all other paths have branch off. Why the tool decides to insert such large delays so early in the path, I do not know.

### DIFF
--- a/flow/vivado_hook_place_design_pre.tcl
+++ b/flow/vivado_hook_place_design_pre.tcl
@@ -1,0 +1,33 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+
+# Manually add delay to the lcd_copi output path. This is to avoid occasional
+# problems when the tool performs all the hold-fixing itself using giant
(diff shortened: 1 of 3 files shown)


marnovandermaas (commented):
I've added the post 1.0 label because I have not run into issues with timing on main. If we do run into this, we can still get it in.

GregAC:
Nice work @elliotb-lowrisc! I think I agree with @marnovandermaas that we're best putting this in after 1.0 but happy to include for 1.0 if you feel strongly we should @elliotb-lowrisc

elliotb-lowrisc:
I think it can wait until after 1.0, assuming the problem doesn't get in the way of release.

GregAC:
With 1.0 out shall we get this merged? @elliotb-lowrisc @marnovandermaas

GregAC (approved):

PR #335, opened 2024-11-07
Doc updates

marnovandermaas:


GregAC (approved):
Thanks @marnovandermaas 

We should also add something to explain our special purpose usage of certain SPI CS lines (i.e. for the LCD stuff and the ethernet reset).

AlexJones0 (changes requested):
Can you also change the Raspberry Pi HAT description to say "The Raspberry Pi HAT header has ~27~ **28** pins that can act as GPIO.", as that is still wrong. Also changing the "g27 - g0" to "GPIO 27 to 0" in that table to match the Ardunio Shield description below it might be nice.

Also, the Arduino Shield should not say it has 17 pins but instead 14 pins, from bit offsets 13-0.

It would also be good to note on the PWM docs that even though the registers are 32 bits, only 8 bit counters are implemented and as such any writes larger than 8 bits are invalid.

marnovandermaas:
> Can you also change the Raspberry Pi HAT description to say "The Raspberry Pi HAT header has ~27~ **28** pins that can act as GPIO.", as that is still wrong. Also changing the "g27 - g0" to "GPIO 27 to 0" in that table to match the Ardunio Shield description below it might be nice.
> 
> Also, the Arduino Shield should not say it has 17 pins but instead 14 pins, from bit offsets 13-0.
> 
> Might also be nice to note next to the `0x000C` output enable register that it is unused/meaningless.
> 
> It would also be good to note on the PWM docs that even though the registers are 32 bits, only 8 bit counters are implemented and as such any writes larger than 8 bits are invalid.
> 
> SPI block also still says it has 64 bit FIFOs, which I believe is wrong, as I think it is now just 8?

Thanks, should all be addressed now.

doc/ip/spi.md
```diff
@@ -7,28 +7,26 @@ In Sonata, there are multiple uses for SPI:
...
+| SPI0                  | 0x2000                 |
+| SPI1                  | 0x3000                 |
+| SPI2                  | 0x4000                 |
+
+Please refer to the [pin multiplexer](pinmux/pin-mappings.md) on how to connect SPI 0, 1 and 2.
 
 ## Overview
 
-Each SPI block has a two 64 entry FIFOs one for transmit and one for receive.
+Each SPI block has a two 8 entry FIFOs one for transmit and one for receive.
```
elliotb-lowrisc, line 29: Nit: extra "a" and some punctuation between "FIFOs" and "one" may be called for.
Suggestion: "Each SPI block has two 8 entry FIFOs - one for transmit and one for receive."

marnovandermaas: Thanks, I've taken your suggestion.

PR #336, opened 2024-11-07
Fixup RS485 after pinmux changes

GregAC:


HU90m (approved):

GregAC:
> Yes, it would be nice to get top generator to generate this in the future.

Definitely I think I said something about opening an issue or two for new pinmux features that would make the rs485 integration more elegant I'll add this to the list of things to look at.

PR #337, opened 2024-11-07
Application flash has moved; update spi_test

alees24:
SPI controller 2 (now once again through pinmux) drives the application flash; update spi_test to use the correct controller.
Tested and once again producing correct output.


elliotb-lowrisc (approved):
Works when I test it

PR #338, opened 2024-11-07
Update software for 1.0 release

AlexJones0:
This PR aims to update as much of the software in `sonata-system` as possible so that it will work with the 1.0 release hardware configuration. Note that this does not address:
 - Documentation, which was covered by #335.
 - Fixing `checks/lcd_check`, which was covered by #305. It does however modify this to now use the updated PWM driver, removing a `TODO`.
 - Fixing the legacy hello world and Ethernet demos, which were already covered by #322.
 - Fixing the RS485 checks, which was already covered by #336.
 - Fixing `checks/spi_test`, which was already covered by #337.

See the commit messages for much more detailed explanations.

Note: I have only actually tested some of these changes. I have ran `tests/test_runner`, `checks/rgbled_test`, `checks/gpio_check`, `checks/system_info_check`, and the manual pinmux `checks/pinmux_all_blocks` check all work, but have not tested any of the other changes. Please do test these and let me know if they still aren't working, so that I can aim to fix them.

This PR:
- Bumps the CHERIoT RTOS version being used to the latest revision of the `sonata` branch
- Accordingly updates the software to use the new drivers. 
- It then proceeds to update the common/legacy software definitions to match the latest number, order, configuration and mapping of the various blocks/devices. 
- It updates the legacy I2C driver to match the updated version already fixed in upstream CHERIoT RTOS. 
- It makes changes to the GPIO bit values being used to match the new general GPIO input and output mappings. 
- It fixes any demos that used the LCD previously to now make use of the chip selects and backlight PWM in place of the GPIO pins.
- It fixes the pinmux checker to account for the fact that a third SPI is now muxable.

Again, please do run tests to catch anything that this does not yet fix.


marnovandermaas (commented):
Looks good from code review, running tests now.

marnovandermaas (approved):
All checks are passing for me. I added a fix for the revocation test in a commit on top of this as well.

AlexJones0:
Your latest commit looks good to me if it makes the test pass, thanks for adding that.

alees24:
I'm happy with the software changes from reading through them; just a couple of observations but I'll leave it to @marnovandermaas to approve after testing.
CLINT_ADDRESS/BOUNDS are a little confusion and could perhaps benefit at some point from an annotation that this is for the timer.
We should probably at some point have a list of IRQ numbersI'm happy with the software changes from reading through them; just a couple of observations but I'll leave it to @marnovandermaas to approve after testing.
CLINT_ADDRESS/BOUNDS are a little confusion and could perhaps benefit at some point from an annotation that this is for the timer.
We should probably at some point have a list of IRQ numbers available to the legacy sw world too, not just those in the CHERI world PLIC, so that e.g. we can avoid hardcoded `i2c0_irq`

sw/cheri/boot/boot_loader.cc
```diff
@@ -148,9 +148,9 @@ extern "C" uint32_t rom_loader_entry(void *rwRoot) {
 
   SpiPtr spi = spi_ptr(root, 2);
 
-  CHERI::Capability<volatile SonataGPIO> gpio = root.cast<volatile SonataGPIO>();
-  gpio.address()                              = GPIO_ADDRESS;
-  gpio.bounds()                               = GPIO_BOUNDS;
+  CHERI::Capability<volatile SonataGpioBoard> gpio = root.cast<volatile SonataGpioBoard>();
+  gpio.address()                                   = GPIO_ADDRESS;
+  gpio.bounds()                                    = GPIO_BOUNDS;
```
HU90m, line 153: Would be nice to use the 'sonata-devices' functions.

HU90m (approved):

PR #339, opened 2024-11-07
Enable SPI external loopback test between MB3 & MB4

AlexJones0:
Courtesy of a request from @alees24. Enables the (already implemented) SPI external loopback test by adding an external jumper cable between pins MB3 and MB4 on the mikroBUS click header. This has been added to the FPGA in CI so that this should pass. Performs appropriate pinmuxing if the SPI test is configured to run this test, so that the test can pass.


alees24 (approved):
Thanks.

PR #341, opened 2024-11-08
Connect arduino shield pin 9 to correct PWM channel

GregAC:
Previously both pin 6 and 9 were connected to PWM channel 2


marnovandermaas (approved):
Looks reasonable to me.

Issue #228, opened 2024-10-10
SRAM address aliasing when HyperRAM is enabled


djlowther:
When HyperRAM is enabled, the the SRAM address bus is only 17 bits wide but  `ADDR_MASK_SRAM` is still `32'h 0003ffff`, leading to reads from/writes to addresses 0x00120000-0x0013ffff being accepted but accessing the same physical memory as 0x00100000-0x0011ffff.

marnovandermaas:
Yes, we need to change the size of the SRAM to `0x00020000` when HyperRAM is disabled here: https://github.com/lowRISC/sonata-system/blob/e73c4574b44dfaabbf57d2418beb3e1fca86e240/data/xbar_main.hjson#L38
We probably need to remove the automatically generated files and get them to be generated by FuseSoC based on the DisableHyperRam flag: https://github.com/lowRISC/sonata-system/blob/e73c4574b44dfaabbf57d2418beb3e1fca86e240/sonata.core#L87

PR #345, opened 2024-11-11
Memory map updated for v1.0

marnovandermaas:
7 MiB of HyperRAM is not accessible yet, so is marked as reserved instead. Also there is now only 128 KiB of SRAM. HyperRAM control and debug module are not available either so these are just marked as reserved for now.

### DIFF
--- a/doc/dev/ref-manual.md
+++ b/doc/dev/ref-manual.md
@@ -35,15 +35,15 @@ Wherever possible, we reuse existing, high-quality, open-source hardware IP bloc
 For all registers in this section, the functionality is mapped onto the least significant bits of registers and each register is 32 bits wide.
 
 | Base address |    Size | Functionality        |
-|--------------|--------:|----------------------|
-| 0x0010_0000  | 256 KiB | Internal SRAM        |
-| 0x3000_0000  |   4 KiB | Revocation tags      |
-| 0x4000_0000  |   1 MiB | Tagged RAM           |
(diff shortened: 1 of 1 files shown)


doc/dev/ref-manual.md
```diff
@@ -35,15 +35,15 @@ Wherever possible, we reuse existing, high-quality, open-source hardware IP bloc
 For all registers in this section, the functionality is mapped onto the least significant bits of registers and each register is 32 bits wide.
 
 | Base address |    Size | Functionality        |
-|--------------|--------:|----------------------|
-| 0x0010_0000  | 256 KiB | Internal SRAM        |
+|--------------|---------|----------------------|
+| 0x0010_0000  | 128 KiB | Internal SRAM        |
```
elliotb-lowrisc, line 39: xbar_main.hjson seems to have the old value?

marnovandermaas: Yes, we can still increase to 256 KiB of SRAM when HyperRAM is disabled: https://github.com/lowRISC/sonata-system/blob/2700cbb6d335188c3d1058f1cdd8319578d65269/rtl/system/sonata_system.sv#L110
However HyperRAM is enabled by default and I want the documentation to show the default configuration.

marnovandermaas: Resolving this is tracked in this issue by the way: https://github.com/lowRISC/sonata-system/issues/228

elliotb-lowrisc (approved):
Ah right, review yes

PR #347, opened 2024-11-12
HyperRAM permanently enabled

marnovandermaas:
SRAM is now fixed at 128 KiB

HyperRAM can no longer be disabled. This stops our memory map needing to keep space for 256 KiB of SRAM even though our default configuration only uses 128 KiB.

This commit also fixes the size of the revocation tags.


elliotb-lowrisc:
FPGA build seems good to me (and seems to work)

dv/verilator/top_verilator.sv
```diff
@@ -301,8 +299,7 @@ module top_verilator (input logic clk_i, rst_ni);
   sonata_system #(
     .CheriErrWidth   ( CheriErrWidth   ),
     .SysClkFreq      ( SysClkFreq      ),
-    .HRClkFreq       ( HRClkFreq       ),
-    .DisableHyperram ( DisableHyperram )
+    .HRClkFreq       ( HRClkFreq       )
```
HU90m, line 302: While you're editing these lines, would you mind addressing https://github.com/lowRISC/sonata-system/pull/149#discussion_r1718306837 ?

HU90m (approved):
Brillant! Thanks marno

rtl/system/sonata_system.sv
```diff
@@ -668,8 +641,8 @@ module sonata_system
     end
   end
 
-  // Size of revocation tag memory is 4 KiB, one bit for each 64 in SRAM
-  localparam int unsigned RevTagDepth = 4 * 1024 * 8 / BusDataWidth;
+  // Size of revocation tag memory is one bit for each 64 in SRAM.
+  localparam int unsigned RevTagDepth = MemSize / 64 * 8 / BusDataWidth;
```
GregAC, line 645:
```suggestion
  localparam int unsigned RevTagDepth = ((MemSize * 8) / 64) / BusDataWidth;
```

Or similar would make more sense given the comment but perhaps don't convert up to bits anyway and just say '1 bit of revocation tag memory per 8-bytes (64 bits) in SRAM' with:

```
  localparam int unsigned RevTagDepth = (MemSize / 8) / BusDataWidth;
```

Parenthesis aren't required but I prefer to make expressions obvious with extra brackets.

GregAC (approved):
Minor potential tweak around the memory depth parameter but otherwise looks good, thanks for sorting this out.

dv/verilator/top_verilator.sv
marnovandermaas: I've now updated that, thanks for the reminder!

rtl/system/sonata_system.sv
marnovandermaas: Yes, that looks much nicer. Now updated.

PR #351, opened 2024-11-12
DRAFT: RTOS test suite on FPGA runner

marnovandermaas:


marnovandermaas:
The reason this is failing is because we need a file system mounted to copy the UF2 file to the FPGA. We may be able to work around this by using `dd` to the device directly.

HU90m:
> The reason this is failing is because we need a file system mounted to copy the UF2 file to the FPGA. We may be able to work around this by using `dd` to the device directly.

Yeah, @nbdd0121 , @engdoreis and have discussed possible solutions to this. I think it would be nice to load the RTOS in the same way we load tests. This would require either overwriting the boot loader so we can load normal RTOS images (i.e. normal start address). Two straight forward solutions:
1. Use vmems and  bung a simple bootloader stub atop the RTOS binary.
2. Use bitstream splicing to change the firmware.

As far as I'm aware, this is still on @engdoreis todo list

PR #354, opened 2024-11-13
sw: Refactored pinmux driver

HU90m:
The changes allow for easier access control and compartmentalisation.

Also the pinmux block input documentation names to match the driver. *The names in the driver can't have square brackets for IO indices.*


sw/cheri/common/platform-pinmux.hh.tpl
```diff
@@ -1,195 +1,201 @@
...
+using Debug = ConditionalDebug<DebugDriver, "Pinmux">;
+
+/**
+ * Each pin sink is configured by an 8-bit register. This enum maps pin sink
+ * names to the offset of their configuration registers. The offsets are relative
+ * to the first pin sink register.
+ *
+ * Documentation sources:
+ * 1. https://lowrisc.github.io/sonata-system/doc/ip/pinmux/
+ * 2. https://github.com/lowRISC/sonata-system/blob/4b72d8c07c727846c6ccb27754352388f3b2ac9a/data/pins_sonata.xdc
```
AlexJones0, line 30: Might be nice to bump this permalink as it is a bit out of date, and doesn't include e.g. the PMODC or RS485 pins and a couple of other changes.

sw/cheri/common/platform-pinmux.hh.tpl
```diff
@@ -1,195 +1,201 @@
...
-    registers[registerOffset] = (1 << option);
-    return true;
-  }
+/**
+ * A handle to a sink configuration register. This can be used to select
+ * the source of the handle's associated sink.
+ */
+template <typename SinkEnum>
+struct Sink {
+  CHERI::Capability<volatile uint8_t> reg;
```
AlexJones0, line 120: Out of curiosity, how do you plan to handle this in the RTOS version of the driver? Currently I think you'd have to get a capability to the entire Pinmux range and then modify that to only the range of the specific sink selector register, and so while the compartmentalisation would be nicer, auditing would still show access to the entire Pinmux MMIO region even if only a small subset is being used.

sw/cheri/common/platform-pinmux.hh.tpl
```diff
@@ -1,195 +1,201 @@
...
-   * capability to the pinmux registers. This should be replaced with
-   * an appropriate `MMIO_CAPABILITY` call in the version of the driver
-   * that runs in CHERIoT RTOS, rather than baremetal.
-   */
-  SonataPinmux(volatile uint8_t *registers) : registers(registers) {}
+  /// Disconnect the sink from all available sources.
+  void disable() { *reg = 0b01; }
+
+  /// Reset the sink to it's default source.
+  void default_selection() { *reg = 0b10; }
```
AlexJones0, line 146: Nit: I think I prefer using `(1 << 0)` and `(1 << 1)` in place of the literals `0b01` and `0b10` just so it's clearer to anyone reading how the `disable` and `default_selection` methods are related to the `select` method above, but that's up to your preference.

AlexJones0 (approved):
This looks like a big improvement to me, thanks for working on this @HU90m! Just a few comments. 

I've tested with the manual pinmux check procedure and that still works fine. The simple pinmux check is broken however, due to some slightly incorrect logic when I wrote that (I wait for the transmit FIFO to be empty, but not necessarily for transmission to finish). The change from using `pinmux.select...()` to `sink.disable()` actually optimises the pinmux enough that it causes this bug to manifest. I fixed it with the following patch:

```patch
diff --git a/sw/cheri/checks/pinmux_check.cc b/sw/cheri/checks/pinmux_check.cc
index e6d71535..be8cd027 100644
--- a/sw/cheri/checks/pinmux_check.cc
+++ b/sw/cheri/checks/pinmux_check.cc
@@ -25,9 +25,7 @@ using namespace CHERI;
  * Blocks until the UART transmit FIFO is empty.
  */
 void block_until_uart_tx_done(Capability<volatile OpenTitanUart> uart) {
-  while (uart->transmit_fifo_level() > 0) {
-    asm volatile("");
-  }
+  while (!(uart->status & uart->StatusTransmitIdle));
 }
 
 /**
```

sw/cheri/common/platform-pinmux.hh.tpl
```diff
@@ -1,195 +1,201 @@
...
+};
+
+namespace {
+template <typename SinkEnum>
+// This is used by `BlockSinks` and `PinSinks`
+// to return a capability to a single sink's configuration register.
+inline Sink<SinkEnum> _get_sink(volatile uint8_t *base_register, const SinkEnum sink) {
+  CHERI::Capability reg = {base_register + static_cast<ptrdiff_t>(sink)};
+  reg.bounds()          = sizeof(uint8_t);
+  return Sink<SinkEnum>{reg, sink};
```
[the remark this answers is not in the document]
HU90m: There will be two contiguous MMIO regions described in the board file: block sinks and pin sinks. Any compartment could request access to either or both. This can be audited normally.

Yes, the idea of having `{Pin,Block}Sinks::get()` return a capability with a bound only covering the particular sink is that this allows a compartment to give another compartment permission to configure the particular sink, without giving access to the whole region.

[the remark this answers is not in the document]
AlexJones0: I see, that makes sense. I think it would be nicer if for auditing purposes no compartments were ever made to get access to the full MMIO regions, and instead only had to get access to the pins required, but I suppose that would either require defining 155 sink devices in the board description file, or (preferably) changes in upstream `cheriot-rtos` to support this kind of granular capability loading functionality (e.g. `MMIO_CAPABILITY(uint8_t, pinmux, offset=0x012, bounds=0x8)` would be nice for example).

HU90m:
Thanks for the speedy and thorough review Alex!

PR #355, opened 2024-11-15
Block diagram includes pinmux and other updates

marnovandermaas:
For example SRAM is 128 KiB, there's no DMA, there's no CHERIoT Debug module, etc.

Re-do of this PR: https://github.com/lowRISC/sonata-system/pull/264 addressing the comments of @HU90m.


elliotb-lowrisc (approved):

HU90m (commented):
Nits:
1. The USB device connects to a different USB port. Connecting it to the same hub may cause confusion.
2. The debug module is connected to the crossbar. My suggestion was to reroute it's connection so it isn't going under the USB.

marnovandermaas:
>     1. The USB device connects to a different USB port. Connecting it to the same hub may cause confusion.
I've added a user USB now which should hopefully help this.
>     2. The debug module is connected to the crossbar. My suggestion was to reroute it's connection so it isn't going under the USB.
I actually decided to remove the debug module because it's not really user visible.

PR #358, opened 2024-11-20
Hardware revoker control re-mastered

marnovandermaas:
The hardware revoker was controlled by a memory mapped register from the CHERIoT Safe repository. This PR ports this to use reggen registers. This will make the IP more portable, specifically for the Sunburst chip repository.


rtl/ip/rev_ctl/rtl/rev_ctl.sv
```diff
@@ -0,0 +1,93 @@
...
+
+  assign ctl_to_core_o = {63'h0, go, reg2hw.end_addr.q, reg2hw.start_addr.q};
+  assign rev_ctl_irq_o = reg2hw.interrupt_status.q & reg2hw.interrupt_enable.q;
+
+  // Hardware to registers.
+  // Read values for hardware extended registers.
+  assign hw2reg.go.d = 32'h5500_0000;
+  assign hw2reg.interrupt_status.d = interrupt_status;
+  // Increment epoch when done with a sweep.
+  assign hw2reg.epoch.epoch.d = reg2hw.epoch.epoch.q + (done ? 31'b1 : 31'b0);
```
GregAC, line 74: Nit: no need to factor `done` in here as the enable term only writes if `done` is set anyway.

```suggestion
  assign hw2reg.epoch.epoch.d = reg2hw.epoch.epoch.q + 31'b1;
```

sw/cheri/checks/tag_test.S
```diff
@@ -138,6 +138,28 @@ revtags:
...
+  addi t2, t2, -8 // Point to last capability aligned region in SRAM
+  csw t2, 4(ct1) // Set end of memory to sweep
+  addi t2, t2, -8 // Second to last capability in SRAM.
+  csw t2, 0(ct1) // Set start of memory to sweep
+  clw t2, 12(ct1) // Read current epoch
+  andi t2, t2, 1 // Get running bit
+  bnez t2, fail
+  li t2, 1
+  csw t2, 20(ct1) // Enable interrupts
+  csw t2, 8(ct1) // Start sweep
```
GregAC, line 161: Shouldn't we extend the test to ensure the revoker finishes and setup something for it to revoke? As is we're testing it doesn't report it's running if we haven't started it and that we're not immediately deadlocking the CPU by starting it off.

GregAC (approved):
Thanks @marnovandermaas 

We can address the potential testing extension in a follow up PR. For sunburst-chip we'll need to do a more in-depth test anyway so perhaps not worth spending time on this now.

sw/cheri/checks/tag_test.S
marnovandermaas: Yes, at the moment I am running this test in simulation and checking the waveform to see if it does what I expect. Let me see if I can extend this a bit so that it at least tests something in case no-one looks at the waveform.

rtl/ip/rev_ctl/rtl/rev_ctl.sv
marnovandermaas: Ah, good point! I'll make that change.

sw/cheri/checks/tag_test.S
marnovandermaas: I've at least checked the interrupt status now, but I agree we should probably add an actual revocation in here.

PR #359, opened 2024-11-28
Simulation documentation: add Docker instructions

marnovandermaas:
The code was tested as part of this PR: https://github.com/CHERIoT-Platform/devcontainer/pull/19


elliotb-lowrisc (approved):
I've no idea what a Dockerfile should look like, but if it's for advances users then it is probably better to just get it in rather than languishing as a PR

PR #360, opened 2024-11-28
SD card access and Read Only FAT32 file system support

alees24:
**CI failures are expected**; there is no microSD card( image) in ~~either~~ simulation~~or FPGA testing~~ at present.

This PR does the following:

Commit 1:
- Introduces SD card support on top of the SPI driver
- Adds a _manual_ 'check' program called 'sdraw_check' that insists upon manual insertion of a card that may be corrupted by the test, and then performs a number of randomised block writes and reads to check that these occur successfully.

Commit 2:
- Introduces Read Only support for a single FAT32 partition (as SD cards are typically shipped by manufacturers).
- Adds SD card testing to the 'test_runner' regression test in CI to test (i) SD card presence, (ii) the ability to read the card properties, (iii) the ability to read a sample text file from the root directory of a FAT32 partition on the SD card (iv) that it matches against a reference copy.

Whilst 'sdraw_check' may be run against any modern microSD card of suitable capacity _that does not contain data you wish to keep_, the test program in the second commit requires that the card be prepared with a suitable file 'LOREM.IPS' (with that exact name and capitalisation at present). Detailed instructions may be found in the source file `sw/cheri/tests/sdcard_tests.hh`; in outline, modify that file by setting 'emitText = true' before running it for the first time and it will emit the instructions and the sample text that must be used in setting up the microSD card.

Create a single FAT32 partition on the card, using the 'Master Boot Record' scheme and not 'GPT.' Newly-purchased cards should already be suitable although only 32GiB and 64GiB cards have been tried to date.

Still to come, and probably required before we can merge this (to keep the FPGA and simulation consistent):
~~1. DPI model of a SPI-connected SD card; this allows the above test to pass in simulation too.~~

Still on the 'TODO' list - but perhaps may be left - until a later PR are:
~~1. Improved filename matching within directories, including Long FileName support.~~
~~2. A fresh attempt to get multiple block read/write operations working; these are presently implemented using repeated single block reads.~~

### DIFF
--- a/sw/cheri/checks/sdraw_check.cc
+++ b/sw/cheri/checks/sdraw_check.cc
@@ -1,0 +1,177 @@
+/**
+ * Copyright lowRISC contributors.
+ * Licensed under the Apache License, Version 2.0, see LICENSE for details.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * SD card raw write/read block test; this is a MANUAL check that should be performed only
+ * on a card THAT DOES NOT CONTAIN ANY DATA OF VALUE.
+ *
+ * It does not require formatting or any particular file system to be present.
+ *
+ * If there is a card inserted when the test starts up, you will be asked to remove the
+ * card and then insert it after the warning message is displayed; this is to decrease the
+ * likelihood of inadvertent data loss.
+ */
+
+#define CHERIOT_NO_AMBIENT_MALLOC
+#define CHERIOT_NO_NEW_DELETE
+#define CHERIOT_PLATFORM_CUSTOM_UART
+
+#include "../common/sonata-devices.hh"
+// clang-format off
+#include <ds/xoroshiro.h>
+#include <cheri.hh>
+// clang-format on
+#include "../common/console.hh"
+#include "../common/sdcard-utils.hh"
+#include "../common/uart-utils.hh"
+
+// Maximum number of blocks per transfer; reasonable model of FS activity.
+static constexpr unsigned kMaxBlocks = 0x10u;
+// Must be a power-of-two.
+static_assert(!(kMaxBlocks & (kMaxBlocks - 1u)));
+
+static constexpr unsigned kBlockLen = 0x200u;
+
+// Diagnostic logging?
+static const bool kLogging = false;
+
+static uint8_t write_data[kMaxBlocks * kBlockLen];
+static uint8_t read_data[kMaxBlocks * kBlockLen];
+static uint8_t orig_data[kMaxBlocks * kBlockLen];
+
+// Compare a sequence of bytes against a reference, returning the number of mismatches.
+static int compare_bytes(const uint8_t *ref, const uint8_t *data, size_t len, Log &log) {
+  unsigned mismatches = 0u;
+  while (len-- > 0u) {
+    mismatches += *ref++ != *data++;
+  }
+  return mismatches;
+}
+
+[[noreturn]] extern "C" void entry_point(void *rwRoot) {
+  CapRoot root{rwRoot};
+
+  auto uart0 = uart_ptr(root);
+  uart0->init(BAUD_RATE);
+  WriteUart uart{uart0};
+  Log log(uart);
+
+  // The SPI controller talkes to the microSD card in SPI mode.
+  auto spi = spi_ptr(root, 2);
+
+  // We need to use the pinmux to select the microSD card for SPI controller 2 reads (CIPO),
+  // as well as preventing outbound traffic to the microSD card also reaching the application
+  // flash (for safety; it _should_ ignore traffic not accompanied by Chip Select assertion).
+  auto pin_output                 = pin_sinks_ptr(root);
+  SonataPinmux::Sink appspi_cs    = pin_output->get(SonataPinmux::PinSink::appspi_cs);
+  SonataPinmux::Sink appspi_clk   = pin_output->get(SonataPinmux::PinSink::appspi_cs);
+  SonataPinmux::Sink appspi_d0    = pin_output->get(SonataPinmux::PinSink::appspi_d0);
+  SonataPinmux::Sink microsd_dat3 = pin_output->get(SonataPinmux::PinSink::microsd_dat3);
+  SonataPinmux::Sink microsd_clk  = pin_output->get(SonataPinmux::PinSink::microsd_clk);
+  SonataPinmux::Sink microsd_cmd  = pin_output->get(SonataPinmux::PinSink::microsd_cmd);
+
+  auto block_input              = block_sinks_ptr(root);
+  SonataPinmux::Sink spi_0_cipo = block_input->get(SonataPinmux::BlockSink::spi_0_cipo);
+
+  // Suppress traffic to the application flash.
+  appspi_cs.disable();
+  appspi_clk.disable();
+  appspi_d0.disable();
+  // Direct SPI controller 2 to drive the microSD pins.
+  microsd_dat3.default_selection();
+  microsd_clk.default_selection();
+  microsd_cmd.default_selection();
+  // Select microSD CIPO as SPI controller input.
+  constexpr uint8_t PmuxSpi0CipoToSdDat0 = 2;
+  spi_0_cipo.select(PmuxSpi0CipoToSdDat0);
+
+  // We need to use the GPIO to detect card presence.
+  auto gpio = gpio_ptr(root);
+
+  // microSD card is on Chip Select 1 (0 goes to the application flash).
+  constexpr unsigned csBit = 1u;
+  // microSD card detection bit is on input 16.
+  constexpr unsigned detBit = 16u;
+
+  log.println("SD card raw write/read test.");
+
+  SdCard sd(spi, gpio, csBit, detBit, true);
+
+  // Wait until there is no card in the slot before issuing a warning message.
+  if (sd.present()) {
+    log.println("Please remove the microSD card from the slot.");
+    while (sd.present()) {
+      // Prevent erroneous optimisation to infinite loop (.l: j l)
+      asm("");
+    }
+  }
+  // Wait until a card is detected.
+  if (!sd.present()) {
+    log.println("Please insert a microSD card that does not contain any valued data.");
+    log.println("*** DATA BLOCKS WILL BE OVERWRITTEN ***");
+    while (!sd.present()) {
+      // Prevent erroneous optimisation to infinite loop (.l: j l)
+      asm("");
+    }
+  }
+
+  log.println("Starting write/read test.... ");
+
+  sd.init();
+
+  // Initialise random number generation.
+  ds::xoroshiro::P64R32 prng;
+  prng.set_state(0xDEADBEEF, 0xBAADCAFE);
+
+  const unsigned numOps = 0x10u;
+  int failures          = 0;
+  for (unsigned b = 0u; !failures && b < numOps; b++) {
+    // Choose a random initial block.
+    uint32_t blk = (uint16_t)prng();
+    // Choose a random transfer length.
+    unsigned num = 1 + ((kMaxBlocks - 1u) & prng());
+    log.println("Testing {} block(s) from block {} onwards", num, blk);
+    // Collect the data from those blocks.
+    log.println("Reading original data...");
+    failures += !sd.read_blocks(blk, orig_data, num);
+    if (!failures) {
+      if (kLogging) {
+        log.println("Original data:");
+        dump_bytes(log, orig_data, num * kBlockLen);
+      }
+      // Randomise the data that we're going to write.
+      for (unsigned idx = 0u; idx < num * kBlockLen; idx++) {
+        write_data[idx] = (uint8_t)prng();
+      }
+      if (kLogging) {
+        log.println("Write data:");
+        dump_bytes(log, write_data, num * kBlockLen);
+      }
+      log.println("Writing...");
+      // Write out the data to the chosen block.
+      failures += !sd.write_blocks(blk, write_data, num);
+      if (!failures) {
+        log.println("Reading back...");
+        // Read it back.
+        failures += !sd.read_blocks(blk, read_data, num);
+        if (kLogging) {
+          log.println("Read data:");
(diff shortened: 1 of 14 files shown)


sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,389 @@
...
+    spi->blocking_read(buf, len);
+    spi->blocking_read(crc16, sizeof(crc16));
+
+    select_card(false);
+
+    return true;
+  }
+
+  // Read Card Specific Data (CSD).
+  bool read_csd(uint8_t *buf, size_t len) {
```
marnovandermaas, line 151: This is nearly the same as the CID read. Can you avoid the code repetition?

sw/cheri/tests/sdcard_tests.hh
```diff
@@ -0,0 +1,291 @@
...
+static constexpr bool logging = false;
+
+// Set this to true to emit the `lorem ipsum` sample text for capture and subsequent
+// writing to a FAT32-formatted microSD card as `LOREM.IPS` within the root directory.
+static constexpr bool emitText = false;
+
+static uint8_t fileBuffer[BLOCK_LEN];
+
+// Dump out a sequence of bytes as hexadecimal and ASCII text.
+static void dump_bytes(const uint8_t *buf, size_t blkBytes, Log &log) {
```
marnovandermaas, line 50: Is this the same as the one in the sdcard check? Can we re-use them somehow?

sw/cheri/tests/sdcard_tests.hh
```diff
@@ -0,0 +1,291 @@
...
+  gpio.address()                            = GPIO_ADDRESS;
+  gpio.bounds()                             = GPIO_BOUNDS;
+
+  // microSD card is on Chip Select 1 (0 goes to the application flash).
+  constexpr unsigned csBit = 1u;
+  // microSD card detection bit is on input 16.
+  constexpr unsigned detBit = 16u;
+
+  // Initialise SD card access, using CRCs on all traffic.
+  SdCard sd(spi, gpio, csBit, detBit, true);  //, &log);
```
marnovandermaas, line 188: Nit: remove the commented out code?

sw/cheri/checks/sdraw_check.cc
```diff
@@ -0,0 +1,169 @@
...
+    // Write out the data to the chosen block.
+    sd.write_blocks(blk, write_data);
+    dump_bytes(write_data, kBlockLen, log);
+    // Read it back.
+    sd.read_blocks(blk, read_data);
+    dump_bytes(read_data, kBlockLen, log);
+    // Check each of the read bytes against what we tried to write.
+    failures += compare_bytes(write_data, read_data, kBlockLen, log);
+  }
+  write_test_result(log, failures);
```
marnovandermaas, line 164: Is this function only used for tests, or can it also be used in a check?

marnovandermaas (commented):
Thanks for putting this together. I just left a few comments while reading through the code. We'll need to get CI to pass before merging, but this is good enough to share around to people that want to get started with SD cards.

GregAC:
Thanks for this @alees24.

I'm keen we do not spend significant time on this right now so let's not worry about spending time on an SD card DPI model. As we already have the test code it's reasonable for us to include the test in CI to run against the real FPGA.

We also need to consider how to to best use this in sonata-software. It's important we have a working example in there, so where shall `sdcard-utils.hh` live? We could hopefully refactor things so it can work both in the bare metal environment and CHERIoT RTOS environment. I think it's best left in this repository so we don't introduce a new sonata-system -> sonata-software dependency. We have a sonata-software -> sonata-system link though it's done via nix. In order to provide easy code access for people who aren't using the nix environment (i.e. we want a way to build an SD card example in sonata-software without relying on some nix magic to pull the `sdcard-utils.hh` into the right place) either we bring sonata-system in as a sub-module or we just make a copy of it/vendor it in sonata-software. I think I favour the latter as it's just one file and we don't really want the whole of sonata-system as a big submodule.

We could also just say no sd card util at all in sonata-system to simplify things. That would mean no testing of sd card functionality in CI though. I don't think it's that important as it's just a plain SPI device but would mean if we mess up something do with the pinmux/pin mapping it would get flagged.

alees24:
The DPI model already exists and works; sdcard_tests.hh works in sim and it's even possible to play video files to the simulated LCD; that was done a couple of weeks ago. I guess it wasn't clear that I meant only that it needs a little tidying before inclusion, so it's not in this PR.

sw/cheri/checks/sdraw_check.cc
```diff
@@ -0,0 +1,169 @@
...
+
+#define CHERIOT_NO_AMBIENT_MALLOC
+#define CHERIOT_NO_NEW_DELETE
+#define CHERIOT_PLATFORM_CUSTOM_UART
+
+#include "../common/sonata-devices.hh"
+// clang-format off
+#include <ds/xoroshiro.h>
+#include <cheri.hh>
+// clang-format on
```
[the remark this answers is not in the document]
alees24: Copied/mimicked code; it's done in other files. I think we even still have a source file that #includes a single file twice, once inside those comments and once outside!

sw/cheri/tests/sdcard_tests.hh
alees24: There are presently 3 copies, and one has a comment about moving this into console.hh; I'll move it there unless you can think of a better home?

GregAC:
> The DPI model already exists and works; 

Awesome. May as well leverage it then.

re: example in sonata-software I think the thing to do is get this PR merged ignoring that issue. We can then deal with that separately.

sw/cheri/tests/sdcard_tests.hh
alees24: I've moved it to console.hh as proposed.

alees24:
Right, I think this is ready to go; long filename support and directory traversal/matching could use some more testing but it works well enough for our present needs.

The DPI model is now present, and although the FPGA system presently has problems, 'test_runner' should pass on this PR when they have been resolved. For simulation, however, we have an issue.....it needs as 'sd.img' file that is ca. 1GiB (see the instructions I've added in 'doc/guide/sdcard-setup.md') which won't be present when the simulation is running, presumably....suggestions welcome, please!

sw/cheri/tests/sdcard_tests.hh
marnovandermaas: Thanks, that sounds good to me.

sw/cheri/tests/sdcard_tests.hh
```diff
@@ -0,0 +1,266 @@
...
+
+  // The SPI controller talkes to the microSD card in SPI mode.
+  auto spi = spi_ptr(root, 2);
+
+  // We need to use the pinmux to select the microSD card for SPI controller 2 reads (CIPO),
+  // as well as preventing outbound traffic to the microSD card also reaching the application
+  // flash (for safety; it _should_ ignore traffic not accompanied by Chip Select assertion).
+  auto pin_output                 = pin_sinks_ptr(root);
+  SonataPinmux::Sink appspi_cs    = pin_output->get(SonataPinmux::PinSink::appspi_cs);
+  SonataPinmux::Sink appspi_clk   = pin_output->get(SonataPinmux::PinSink::appspi_cs);
```
AlexJones0, line 131: Just a small error I noticed looking through, shouldn't cause any issues though. 
```suggestion
  SonataPinmux::Sink appspi_clk   = pin_output->get(SonataPinmux::PinSink::appspi_clk);
```

alees24:
I see a few possible options for resolving the simulator situation:
- put the file 'sd.img' in the repository; some people don't like binary files in repositories, and it may need to be quite large; I don't presently know the minimum capacity volume that the code will accept.
- the SD DPI model could be modified to construct a suitable image if there is not one available, or this could be done as a step in the CI build flow; a short sequence of shell commands ought to suffice, and the reference text would need to be extracted from the header file and written as a file on the mounted virtual volume.
- sdcard_tests.hh operates differently on sim c.f. FPGA and does not access a disk image; I'd prefer to avoid this disparity if possible.

Opinions/thoughts, please?

marnovandermaas:
> I see a few possible options for resolving the simulator situation:
> 
>     * put the file 'sd.img' in the repository; some people don't like binary files in repositories, and it may need to be quite large; I don't presently know the minimum capacity volume that the code will accept.
If we need to go done this route, we can use git large file storage.
> 
>     * the SD DPI model could be modified to construct a suitable image if there is not one available, or this could be done as a step in the CI build flow; a short sequence of shell commands ought to suffice, and the reference text would need to be extracted from the header file and written as a file on the mounted virtual volume.
My initial though is that this is best. Would you be able to have a go at this?
> 
>     * sdcard_tests.hh operates differently on sim c.f. FPGA and does not access a disk image; I'd prefer to avoid this disparity if possible.
> 
> 
> Opinions/thoughts, please?

See my opinions inline.

sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,472 @@
...
+    StopTranToken = 0xfdu,
+  };
+
+  // Indicates whether there is an SD card present in the slot.
+  bool present() const { return !(gpio->debouncedInput & det); }
+
+  void select_card(bool enable) { spi->cs = enable ? (spi->cs & ~cs) : (spi->cs | cs); }
+
+  // Initialise the SD card ready for use.
+  bool init() {
```
pamaury, line 93: In my experience working with SD cards, it's better to actually do the initialization properly because you will find cards in the while that enforce the spec to the letter.

sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,472 @@
...
+    // Apparently we're required to send at least 74 SD CLK cycles with
+    // the device _not_ selected before talking to it.
+    spi->blocking_write(ones, 10);
+    spi->wait_idle();
+
+    select_card(true);
+
+    // Note that this is a very stripped-down card initialisation sequence
+    // that assumes SDHC version 2, so use a more recent microSD card.
+    do {
```
pamaury, line 107: Per the spec, a single command should suffice here, do you really need to loop?

sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,472 @@
...
+      send_command(CMD_GO_IDLE_STATE, 0u);
+    } while (0x01 != get_response_R1());
+
+    send_command(CMD_SEND_IF_COND, 0x1aau);
```
pamaury, line 111: Supporting a v1 card is not actually hard: if the card does not answer this command then it's v1. Maybe at least print an error message and bail out?

sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,472 @@
...
+
+  // Indicates whether there is an SD card present in the slot.
+  bool present() const { return !(gpio->debouncedInput & det); }
+
+  void select_card(bool enable) { spi->cs = enable ? (spi->cs & ~cs) : (spi->cs | cs); }
+
+  // Initialise the SD card ready for use.
+  bool init() {
+    // Every card tried seems to be more than capable of keeping up with 20Mbps.
+    constexpr unsigned kSpiSpeed = 0u;
```
pamaury, line 95: I believe that in theory, you are not supposed to go above 400kHz when initializating the card.

sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,472 @@
...
+      send_command(SD_SEND_OP_COND, 1u << 30);
+    } while (0x01 & get_response_R1());
+
+    if (log) {
+      log->println("Setting block length to {}", kBlockLen);
+    }
+
+    // Read card capacity information.
+    send_command(CMD_READ_OCR, 0);
+    get_response_R3();
```
pamaury, line 135: I believe that once you get a successful answer from CMD58 you can switch to the default clock rate which is 20MHz for SPI mode.

sw/cheri/common/sdcard-utils.hh
```diff
@@ -0,0 +1,472 @@
...
+  }
+
+  // Send a command to the SD card with the supplied 32-bit argument.
+  void send_command(uint8_t cmdCode, uint32_t arg) {
+    uint8_t cmd[6];
+    if (log) {
+      log->println("Sending command {:#04x}", cmdCode);
+    }
+
+    // Apparently we need to clock 8 times before sending the command.
```
pamaury, line 262: This is actually the opposite: you need 8 clock ticks *after commands*. Per the spec:

> After the last SD Memory Card bus transaction, the host is required, to provide 8 (eight) clock
> cycles for the card to complete the operation before shutting down the clock. Following is a list of the
> various bus transactions:
> •A command with no response. 8 clocks after the host command end bit.
> •A command with response. 8 clocks after the card response end bit.
> •A read data transaction. 8 clocks after the end bit of the last data block.
> •A write data transaction. 8 clocks after the CRC status token.

marnovandermaas (approved):
Approving this, but let's wait for CI to pass before merging.

sw/cheri/common/sdcard-utils.hh
alees24: Thanks for all the feedback; the code was originally a quick prototype just to check that we could use the SD card at all before 1.0, and it was based upon rather ad hoc tutorials/code found online, without having a proper specification.

It's only example code, but perhaps - at our leisure - we should shore it up properly; @GregAC plans to improve the robustness and error detection of the code.

alees24:
> Re the SD card image file, I'd rather not have everyone who wants to check out the sonata-system repository need to fetch a massive file. Is there any reason we need such a large file? Alternatively we could store it compressed (I'm assuming as an empty file system image it'll compress very very well)
> 
> Is there a way to mark an LFS file as 'optional' or something? I.e. a normal clone won't download it and you need another command to actually fetch it.

I remarked earlier that I don't know how small the image can be and still be usable. 1GiB is merely the size I've been using until now, but it may even work fine with as little as 8KiB, if software can create a suitable image. For now it doesn't matter, we've just restricted the testing in simulation; to be addressed later. It's still exercising the SPI and microSD card interface and doing data reads from CID and CSD in sim, just not from simulated flash storage.

PR #361, opened 2024-11-29
Simulator boot stub for SRAM added

marnovandermaas:
This boot stub is useful for SRAM only examples like the ones in the current upstream RTOS.

### DIFF
--- a/sw/cheri/sim_boot_stub/boot_sram.S
+++ b/sw/cheri/sim_boot_stub/boot_sram.S
@@ -1,0 +1,20 @@
+# Copyright lowRISC contributors.
+# Licensed under the Apache License, Version 2.0, see LICENSE for details.
+# SPDX-License-Identifier: Apache-2.0
+  .section .text.start, "ax", @progbits
+  .zero 0x80
+
+  .globl start
+  .p2align 2
+  .type start,@function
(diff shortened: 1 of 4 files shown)


sw/cheri/sim_boot_stub/boot_sram.S
```diff
@@ -0,0 +1,16 @@
...
+	.section .text.start, "ax", @progbits
+.zero 0x80
+
+	.globl start
+	.p2align 2
+    .type start,@function
+start:
+  // Enable the ICache
+  csrsi            0x7c0, 1
+  auipcc           ct1, 0
```
elliotb-lowrisc, line 13: Perhaps a comment here to explain what it is doing ('jump to ...')

AlexJones0 (approved):
LGTM, but as mentioned might be worth adding a comment and maybe even defining the `0x00101000` immediate (e.g. `#define SRAM_BASE 0x00101000`) so that it is easier to parse what is happening.

sw/cheri/sim_boot_stub/boot_sram.S
marnovandermaas: Thanks, added.

PR #364, opened 2024-12-06
Revocation control test revocation happens.

marnovandermaas:
The test now writes two capabilities to the end of SRAM, one with base zero and one with base at the end of SRAM. After the sweep it checks that the base zero capability's tag stays in tact while the one at the end of SRAM is cleared. Note this is different from the CLC test because the revocation tags are cleared before testing so the CLC will not clear the tags if the hardware revoker hasn't already.

### DIFF
--- a/sw/cheri/checks/tag_test.S
+++ b/sw/cheri/checks/tag_test.S
@@ -140,8 +140,9 @@ revtags:
 
 // Test of hardware revoker
 hw_rev:
-  // Capability to revocation tags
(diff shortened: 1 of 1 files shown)


AlexJones0 (approved):
Thanks @marnovandermaas - especially for the comments, they're very helpful for understanding what the code is doing. I've tested this in simulation and it appears to pass as expected, so it'd be good to get this in and improve the check.

PR #368, opened 2024-12-10
Simulator exit

marnovandermaas:
This PR allows the simulator to exit when the following magic string is sent over the UART:
`Safe to exit simulator.\xd8\xaf\xfb\xa0\xc7\xe1\xa9\xd7`


alees24 (approved):
Is there a particular reason for copying the string into the uartdpi state as opposed to just indexing a constant string; intending to support it being reprogrammed while the DPI model is active?

PR #370, opened 2024-12-19
Add Pinmux PWM test

AlexJones0:
See the commit messages for more detailed explanations. This PR introduces a new `pinmux_pwm_test` which tests that a PWM instance can be muxed, and that the existing PWM loopback testing works only when the pinmux is configured correctly. If running in CI, this test would be sufficient to catch the issue fixed in #369.

I've tested this on both FPGA and in simulation locally, and the FPGA in CI is already set up with a jumper cable so this test should be passing. I've also added the loopback to the simulation model so this can pass there as well.


AlexJones0:
Force push just addresses the typo in the commit message.

PR #374, opened 2024-12-20
SPI doc updated

marnovandermaas:
Remove sentence on chip selects going through GPIO. Also added the INFO and CS registers.

### DIFF
--- a/doc/ip/spi.md
+++ b/doc/ip/spi.md
@@ -33,7 +33,6 @@ Each SPI block has two 8-entry FIFOs - one for transmit and one for receive.
 To begin an SPI transaction write to the [`START`](#start) register.
 Bytes do not need to be immediately available in the transmit FIFO nor space available in the receive FIFO to begin the transaction.
 The SPI block will only run the clock when its able to proceed.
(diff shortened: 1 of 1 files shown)


doc/ip/spi.md
```diff
@@ -290,3 +291,40 @@ Bytes written here are pushed to the transmit FIFO. If the FIFO is full writes a
...
+### Fields
+
+```wavejson_reg
+[{"name": "CS", "bits": 4, "attr": ["rw"], "rotate": 0}, {"bits": 28}]
+```
+
+|  Bits  |  Type  |  Reset  | Name   | Description              |
+|:------:|:------:|:-------:|:-------|:-------------------------|
+|  31:5  |        |         |        | Reserved                 |
+|  4:0   |   rw   |   0xf   | cs     | If this bit is clear the peripheral is selected for transmit/receive operations. |
```
elliotb-lowrisc, line 330: Description uses singular "bit" for a multi-bit field.

elliotb-lowrisc (approved):
Minor issue, but overall looks nice

doc/ip/spi.md
marnovandermaas: Thanks, fixed.

PR #375, opened 2025-01-02
Added notes on using the RPi header for GPIO.

GrassHopper1977:
Added notes to pinmux and GPIO sections of the manual regarding the impact of the pinmux output selector defaulting on 0b10 and the importance of correctly setting the pinmux before attempting to use the RPi header's GPIO pins as outputs.


GrassHopper1977:
This caught me out, so I think it may be worth mentioning.

marnovandermaas (approved):
Thanks for these changes! Feel free to update more of the documentation.

PR #376, opened 2025-01-06
Pinmux driver style fixes

HU90m:
Changed the pinmux driver template so that it outputs code compliant with the CHERIoT RTOS naming conventions.

*See https://github.com/CHERIoT-Platform/cheriot-rtos/pull/360*


sw/cheri/common/platform-pinmux.hh.tpl
```diff
@@ -1,5 +1,10 @@
-// SPDX-FileCopyrightText: lowRISC contributors
-// SPDX-License-Identifier: Apache-2.0
+/**
+ * SPDX-FileCopyrightText: lowRISC contributors
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * This file is generated by ./util/topgen.py
```
nwf, line 5: `top_gen.py`, I think?

marnovandermaas: Fixed now.

marnovandermaas (approved):
Approved after making the change that Gary spotted.

PR #377, opened 2025-01-06
Revocation control: define done earlier

marnovandermaas:
This was causing undeclared identifier warnings.


elliotb-lowrisc (approved):
Looks good to me. Seems to solve simulation issues elsewhere and still builds a bitstream.

PR #382, opened 2025-01-22
xbar: debug host access to hyperram, system_info

### LINKED ISSUE #381: Debug host access to HyperRAM?
Experimenting with https://github.com/CHERIoT-Platform/cheriot-rtos/pull/425, I have a build of the `01.hello_world` example that works fine in the sonata simulator (as in the devcontainer, build cdeeb9a17869, from today) but fails to get off the ground on the actual board.

Some _extremely tedious_ debugging [^0] has brought me back to very early parts of the RTOS loader, specifically https://github.com/CHERIoT-Platform/cheriot-rtos/blob/52a1ba418013e1dbe55addf7b7fcf66e96bb4347/sdk/core/loader/boot.S#L45-L59.  In particular, `ca1` correctly ends up being a tagged capability to `0x00101d74`, in SRAM, and subsequent bytes, but the word read back by `clw s0, IMAGE_HEADER_LOADER_CODE_START_OFFSET(ca1)` is sourced from `0x40001d74`, in HyperRAM, and that's the beginning of an extremely rapid unintentional disassembly of the edifice that the loader is attempting to build.

It certainly _appears_ as though something about the actual synthesized gateware, as opposed to the simulated gateware, is misdirecting a read intended for SRAM to HyperRAM.  Of note, the _instructions_ being executed are being fetched from HyperRAM (and apparently, from observation, correctly so).

Does that theory hold water?  Would the `01.hello_world` ELF and/or dump files be useful for investigation?

[^0]: In particular, this is early enough that basically nothing is initialized, but the 8 GPIO LEDs sure do work.  So, 8 bits of information per recompile and code reload... let's not say "fast fluxing" of images, but I was happy to have OpenOCD working.

nwf:
With the RTOS mainline gaining some support for code in HyperRAM, it would be nice to be able to load such images in over OpenOCD without needing to write code to do bounce buffering.

While here, also give the debug host access to the system_info peripheral, to provide some more diagnostic information.

Should resolve #381, I think?

### DIFF
--- a/data/xbar_main.hjson
+++ b/data/xbar_main.hjson
@@ -231,6 +231,10 @@
       "usbdev",
       "rv_plic",
     ],
-    dbg_host: ["sram"],
+    dbg_host: [
+      "sram",
+      "hyperram",
+      "system_info",
+    ],
   },
 }
--- a/data/xbar_main_generated.hjson
+++ b/data/xbar_main_generated.hjson
@@ -322,6 +322,10 @@
       "usbdev",
       "rv_plic",
     ],
-    dbg_host: ["sram"],
+    dbg_host: [
+      "sram",
+      "hyperram",
+      "system_info",
+    ],
   },
 }
--- a/rtl/bus/xbar_main.sv
+++ b/rtl/bus/xbar_main.sv
@@ -10,11 +10,13 @@
 //   -> s1n_25
 //     -> sm1_26
 //       -> sram
-//     -> hyperram
+//     -> sm1_27
+//       -> hyperram
 //     -> rev_tag
 //     -> gpio
 //     -> pinmux
-//     -> system_info
+//     -> sm1_28
+//       -> system_info
 //     -> rgbled_ctrl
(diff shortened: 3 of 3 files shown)


marnovandermaas:
Here's a bitstream for this run.
[xbar_debug_host_bit.zip](https://github.com/user-attachments/files/18523502/xbar_debug_host_bit.zip)

nwf:
Thanks!  Using the `openFPGALoader`, I am able to load both Hello World and the RTOS test suite over OpenOCD.

nwf:
Can this be merged so it's not forgotten in the next release?

marnovandermaas:
I'm redoing the debug module at the moment and want to include this there if that is ok.

nwf:
That's fine, I think.  Can I ask for an ETA on the debug unit?

marnovandermaas:
I have the debug module integrated but it seems to hang, so need to debug until that's fixed.

PR #387, opened 2025-03-14
CHERIoT debug module integrated

marnovandermaas:
This integrates the CHERIoT debug module into Sonata.

Memory inspection and register inspection work, as well as setting break points.

When trying the "run" command in GDB and asking it to start from the beginning, I get a remote failure reply 'E0E', but I'm hoping this will be solved when using a CHERI-aware LLDB.


elliotb-lowrisc (commented):
I've not managed to get to the point of testing this yet, but here are a couple of early comments.

doc/ip/dm.md
```diff
@@ -1,4 +1,59 @@
...
+```
+
+## JTAG DPI
+
+We also have support for JTAG DPI so you can debug a simulation.
+
+For example in one terminal you can run:
+```sh
+build/lowrisc_sonata_system_0/sim-verilator/Vtop_verilator -t -E sw/cheri/build/checks/uart_
+check
```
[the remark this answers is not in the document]
marnovandermaas: Thanks made one line.

doc/ip/dm.md
```diff
@@ -1,4 +1,59 @@
 # Debug module
```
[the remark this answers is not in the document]
marnovandermaas: I've removed sw-debug.md as it was not being rendered in the mdbook anyways.

doc/ip/dm.md
```diff
@@ -1,4 +1,58 @@
...
 
 The debug module we use in Sonata supports the [RISC-V debug specification](https://github.com/riscv/riscv-debug-spec/releases/download/ar20231012/riscv-debug-stable.pdf).
```
elliotb-lowrisc, line 3: This is a link to the v1.0 spec, but only v0.13.1 is claimed in the debug module docs: vendor/cheriot_debug_module/README.md

doc/ip/dm.md
```diff
@@ -1,4 +1,58 @@
...
+Then in another terminal you can run:
+```sh
+openocd -f util/verilator-openocd-cfg.tcl
+```
+
+## GDB quick guide
+
+To connect to the OpenOCD instance you should run GDB in a separate terminal while OpenOCD is runing in another:
+```sh
+riscv32-unknown-elf-gdb
```
elliotb-lowrisc, line 30: Perhaps some instruction on how to acquire this uncommon program?

doc/ip/dm.md
```diff
@@ -1,4 +1,58 @@
...
+```
+
+When that starts up you should be greeted with a `(gdb)` prompt.
+To connect to your OpenOCD instance, independent of whether this is on FPGA or Verilator, you should use the following comand:
```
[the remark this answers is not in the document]
marnovandermaas: Thanks, fixed.

doc/ip/dm.md
marnovandermaas: Thanks, added a link to the lowrisc toolchain.

marnovandermaas:
Marking as draft until I can fix the FPGA run issues.

elliotb-lowrisc:
> When trying the "run" command in GDB and asking it to start from the beginning, I get a remote failure reply 'E0E'...

I think I've found the cause of this. I noticed the Verilator sim output an `Illegal instruction (hart 0) at PC 0xb0000348: 0x39052023` error when `run` was issued in GDB. This decodes to a `sw x16 0x380(x10)` instruction. The key part is that `rs2` is set to register **16**, which does not exist for CHERIoT Ibex. The debugger doesn't know this, however, as `misa` is set for RV32I, not RV32E. The `run` command seemed to work when I tried hacking `misa` to show the reverse.

marnovandermaas:
> > When trying the "run" command in GDB and asking it to start from the beginning, I get a remote failure reply 'E0E'...
> 
> I think I've found the cause of this. I noticed the Verilator sim output an `Illegal instruction (hart 0) at PC 0xb0000348: 0x39052023` error when `run` was issued in GDB. This decodes to a `sw x16 0x380(x10)` instruction. The key part is that `rs2` is set to register **16**, which does not exist for CHERIoT Ibex. The debugger doesn't know this, however, as `misa` is set for RV32I, not RV32E. The `run` command seemed to work when I tried hacking `misa` to show the reverse.

Cool, that makes sense!

rtl/system/sonata_system.sv
```diff
@@ -127,8 +127,8 @@ module sonata_system
   localparam int unsigned DataBitsPerMask = BusDataWidth / BusByteEnable;
 
   // Debug functionality is disabled.
```
elliotb-lowrisc, line 129: Perhaps update this comment?

elliotb-lowrisc:
Ah ha! The instruction cache is the culprit. It's not respecting the mutable nature of instructions loaded from the AbstractCmd part of the debug module memory map. Register load Abstract Commands kept ending up returning the value of `misa` (and early CSR load) rather than the value of the intended register. Disabling the icache by changing sw/cheri/common/boot.S:62 seemed to finally make breakpoints work as they should (combined with hacked RV32E `misa` value).

marnovandermaas:
@elliotb-lowrisc here's a fix for the MISA issue: https://github.com/lowRISC/sunburst-chip/pull/68

rtl/system/sonata_system.sv
marnovandermaas: Good point, this is done now.

elliotb-lowrisc (approved):
Works for me

PR #391, opened 2025-03-21
Add Sonata XL physical constraints XDC file

elliotb-lowrisc:
Introduce an XDC file for physical constraints for the Sonata XL development board.
The pin mapping for existing ports is completely different, and new ports have also been added.
These new ports include an RGB LED enable line, and many connections to the new headers.
HyperRAM ports have been removed.

Also add mention in the existing physical XDC file that it is for the SONATA-ONE board.


marnovandermaas (approved):
I'm happy for these to go in. Thanks for creating them!

PR #395, opened 2025-03-31
Move onto upstream CHERIoT RTOS

HU90m:
I've tried to keep the changes minimal, so the software doesn't make the best use of the new API

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,244 +1,15 @@
 {
   "nodes": {
-    "attic": {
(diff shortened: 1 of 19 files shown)


AlexJones0 (approved):
All these changes LGTM, especially if CI is passing. I've run a couple of checks locally (e.g. `pinmux_all_blocks`) and it seems to be working fine for me as well.

PR #397, opened 2025-04-03
Improve documentation of test_runner connections

elliotb-lowrisc:
Add a list of things that need to be connected for test_runner to pass and an annotated image of where to connect them.


marnovandermaas (commented):
Thanks for putting this together. @AlexJones0 can you double check that this is also your understanding?

AlexJones0 (approved):
Thanks @elliotb-lowrisc for adding the detailed diagram - it makes the required connections look a bit cleaner then I think they are in reality :)

Is your diagram missing the MikroBUS P7 MB8 "RX" -> MB9 "TX" connection?

elliotb-lowrisc:
> Is your diagram missing the MikroBUS P7 MB8 "RX" -> MB9 "TX" connection?
Ah yes, thanks for catching that. Should be fixed now.

elliotb-lowrisc:
Also forgot to convert some of the text to a path (to avoid missing font problems)

AlexJones0 (approved):

PR #398, opened 2025-04-10
LCD read HW changes and LCD offset SW test

elliotb-lowrisc:
Add HW ability to read data back from the LCD.
Add SW ability to use that to determine if the LCD controller was mis-configured when LCD was assembled, and apply coordinate offset if required.


elliotb-lowrisc:
clang-format

elliotb-lowrisc:
Update sonata.core to reflect removed Vivado hook

marnovandermaas (commented):
What is the behaviour of the check code if it runs on an older bitstream without the receive line connected? Can we make sure we default to the old screen resolution in that case?

elliotb-lowrisc:
On older bitstreams it will complain and then default to not using an offset, as it did previously.

UART output in that case:
```
LCD check
Warning: Unable to determine LCD offset. Try slower SPI clock.
need_offset: NO
```

elliotb-lowrisc:
@HU90m I couldn't figure out if there was a way to regenerate the pin connection diagram to include the new internal connection. Do you know if there is a way?

marnovandermaas:
> I couldn't figure out if there was a way to regenerate the pin connection diagram to include the new internal connection. Do you know if there is a way?

@elliotb-lowrisc I think I've used this command before:
```sh
❯ dot -Tsvg doc/ip/pinmux/pin_mappings.dot -o doc/ip/pinmux/pin_mappings.svg
```

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +152,225 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+    } else if (result == 0x0e) {
+      // Three out of three agree, 132-wide it sadly be
+      return 1;
+    }
+
+    // Software reset to restore most state to default - particularly CASET
+    buf[0] = srst;
+    interface->gpio_write(interface->handle, false, false);
+    interface->spi_write(interface->handle, &buf[0], 1);
+    interface->timer_delay(120);  // 120ms minimum
```
alees24, line 366: From my reading of the datasheet this delay can be as short as 5ms since we're only sending commands here and not expecting the display to come out of a sleep state. It may be worth trying this experimentally and checking that things still work.

It's the caller's responsibility to ensure that the 120ms delay is observed after its own reset operation before expecting the display to become active.

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +152,225 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+// GM[2:0]='000' (132x162) is incorrect for the 128x160 panel actually present,
+// meaning minor coordinate offsets are needed. The offsets are 2 pixels
+// in the narrow dimension (x if portrait) and 1 pixel in the wide dimension
+// (y if portrait). This is due to how the ST7735 controller maps the
+// contents of the internal frame buffer to display itself. See the
+// (unfortunately error-ridden) ST7735 datasheet for more details.
+//
+// NOTE1: Must be run after a HW or SW reset and before any CASET commands.
+// NOTE2: Does NOT always perform a reset before returning. State may be dirty.
+bool need_offset(Capability<volatile OpenTitanUart> &uart, LCD_Interface *interface) {
```
alees24, line 260: The interface here may be passed as a pointer to const (`const LCD_Interface *`).

alees24 (commented):
Thanks for all this work, investigations and testing. I've left a number of comments but they are all minor changes/observations and most have no impact upon operation/require re-testing. The exception to the latter point would be whether the 5ms delay post-SW reset is okay.

sw/cheri/checks/lcd_check.cc
```diff
@@ -176,34 +397,74 @@ extern "C" [[noreturn]] void entry_point(void *rwRoot) {
...
-  // This is maximum speed (15Mbps presently) which works fine, but we can run at a lower speed
-  // to exercise interrupts/FIFO behaviour more thoroughly.
-  const uint8_t SpiSpeed = 0u;
+  // Run at a slower speed at first so that read operations function correctly.
+  const uint16_t SpiReadSpeed = 0x3u;
+  spi->init(false, false, true, SpiReadSpeed);
+
+  // Set the initial state of the LCD control pins.
+  // The CS line in particular needs to be high (deasserted) for the first
+  // command to work.
```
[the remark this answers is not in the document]
elliotb-lowrisc: It's the deasserted->asserted transition that matters. I'll update the comment

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +152,225 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+      interface->gpio_write(interface->handle, true, false);
+      interface->oe_write(interface->handle, true);
+
+#if 0
+      // Debug printout
+      write_hex8b(uart, l);
+      write_str(uart, ": ");
+      write_hex(uart, (buf[0] << 25) | (buf[1] << 17) | (buf[2] << 9) | (buf[3] << 1));
+      write_str(uart, "\r\n");
+#endif
```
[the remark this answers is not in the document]
elliotb-lowrisc: Good point, that looks nicer

sw/cheri/checks/lcd_check.cc
```diff
@@ -216,7 +477,7 @@ extern "C" [[noreturn]] void entry_point(void *rwRoot) {
     uint16_t logo_x1 = logo_x + img_width - 1u;
 
     // Render the logo.
-    draw_image(spi, logo_x, logo_y, logo_x1, logo_y1, (uint8_t *)lowrisc_logo_native);
+    draw_image(spi, logo_x + off_x, logo_y + off_y, logo_x1 + off_x, logo_y1 + off_y, (uint8_t *)lowrisc_logo_native);
```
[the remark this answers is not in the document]
elliotb-lowrisc: Please blame clang-format, which appears to be mandatory for this repo now

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +152,225 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+      }
+    }
+    interface->gpio_write(interface->handle, true, false);
+
+    // Read back a pixel from the start of the second, third, and fourth lines
+    // of the external frame buffer to determine whether the ST7735 controller
+    // in the LCD assembly is configured for 128x160 or 132x162 (by GM pads).
+    // Pixels are always read back in 18-bit format, regardless of COLMOD.
+    result = 0;
+    for (unsigned l = 1u; l < 4; l++) {
```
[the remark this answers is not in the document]
elliotb-lowrisc: Ah, I see. I'll switch to `ln`

sw/cheri/checks/lcd_check.cc
```diff
@@ -176,34 +397,74 @@ extern "C" [[noreturn]] void entry_point(void *rwRoot) {
...
+  bool do_offset = need_offset(uart, &interface);
+  if (do_offset) {
+    write_str(uart, "need_offset: YES\r\n");
+  } else {
+    write_str(uart, "need_offset: NO\r\n");
+  }
+
+  // This is maximum speed (20Mbps presently) which works fine when
+  // only performing writes, but we could run at a lower speed to
+  // exercise interrupts/FIFO behaviour more thoroughly.
```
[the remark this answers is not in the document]
elliotb-lowrisc: Not originally my comment but I'm happy to improve it (by culling the end of it)

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +152,225 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+
+#if 0
+      // Debug printout
+      write_hex8b(uart, l);
+      write_str(uart, ": ");
+      write_hex(uart, (buf[0] << 25) | (buf[1] << 17) | (buf[2] << 9) | (buf[3] << 1));
+      write_str(uart, "\r\n");
+#endif
+
+      if (buf[1] == (patterns[l - 1] >> 1) && buf[2] == (patterns[l - 1] >> 1) && buf[3] == (patterns[l - 1] >> 1)) {
```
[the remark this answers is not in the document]
elliotb-lowrisc: Good idea. I originally had wrapped them to be more readable but clang-format did away with that. I like your suggestion better though.

sw/cheri/checks/lcd_check.cc
elliotb-lowrisc: I'm uncertain about shortening this. The documentation for the SWRESET command seems to claim that 120 ms is needed both in the case of Sleep In and when Sleep Out + Display On mode before sending the next command. Where are you getting the value of 5 ms from?

elliotb-lowrisc:
Addressed further feedback

sw/cheri/checks/lcd_check.cc
alees24: 9.16 Note 7. The poor quality of the datasheet makes me suspect that the line in the SWRESET documentation is incorrect, being duplicated for the 'Sleep In' and 'Sleep Out' modes.

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +154,232 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+   */
+  uint32_t (*gpio_write)(void *handle, bool cs_high, bool dc_high);
+
+  /**
+   * @brief Set the state of the data line output enable.
+   *
+   * @param oe Enable output if true, otherwise disable output (high-Z).
+   *
+   */
+  uint32_t (*oe_write)(void *handle, bool oe);
```
engdoreis, line 197: I'm not sure if you want to mimic the display_driver, but there's no need for this callback, the output can be enabled inside the `spi_write` and disabled inside the `spi_read`.

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +154,232 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+      interface->spi_write(interface->handle, &buf[0], 1u);
+      // Read 1 dummy byte and 3 actual bytes (offset by a dummy clock cycle)
+      interface->oe_write(interface->handle, false);
+      interface->spi_read(interface->handle, buf, 4u);
+      interface->gpio_write(interface->handle, true, false);
+      interface->oe_write(interface->handle, true);
+
+      if (debug) {
+        // Debug printout of values read from LCD frame buffer
+        write_hex8b(uart, ln);
```
engdoreis, line 335: It's not related to this PR, but a side note, this test would benefit from the `log.print`. See example [here]( https://github.com/lowRISC/sonata-system/blob/34c893b1d13b0c5dbacc46a1865e836fca6c05ef/sw/cheri/checks/pinmux_checker.cc#L158)

sw/cheri/checks/lcd_check.cc
alees24: Fair point. I suspect that in time, with more confidence, we can ditch the retrying code entirely. It would be good to do some more extensive testing of the reliability of the reading from the LCD. Clearly the SPI/SD card combination has testing read functionality extensively, so here I'm talking more about the PCB and the quirky line-sharing of the LCD.

sw/cheri/checks/lcd_check.cc
```diff
@@ -150,6 +154,232 @@ static void run_script(Capability<volatile OpenTitanUart> &uart, SpiLcdCap &spi,
...
+// GM[2:0]='000' (132x162) is incorrect for the 128x160 panel actually present,
+// meaning minor coordinate offsets are needed. The offsets are 2 pixels
+// in the narrow dimension (x if portrait) and 1 pixel in the wide dimension
+// (y if portrait). This is due to how the ST7735 controller maps the
+// contents of the internal frame buffer to display itself. See the
+// (unfortunately error-ridden) ST7735 datasheet for more details.
+//
+// NOTE1: Must be run after a HW or SW reset and before any CASET commands.
+// NOTE2: Does NOT always perform a reset before returning. State may be dirty.
+bool need_offset(Capability<volatile OpenTitanUart> &uart, const LCD_Interface *interface) {
```
engdoreis, line 262: This test already imports the display_driver, so maybe it could use the same function that is defined there?
https://github.com/engdoreis/display_drivers/tree/v0.2.0?tab=readme-ov-file#detecting-whether-offset-is-needed

alees24 (approved):
Thanks for the changes. As discussed directly, there are a couple of remaining comments that could perhaps be refined, but this looks good. Great to have a workaround!

sw/cheri/checks/lcd_check.cc
elliotb-lowrisc: Hmm, good point. Might leave it explicit for now to make it clearer what's going on, seeing as this is something of a reference implementation.

sw/cheri/checks/lcd_check.cc
elliotb-lowrisc: I couldn't get that to work straight away so will leave it for later so we can get this in before the end of the day (and I go on holiday)

sw/cheri/checks/lcd_check.cc
elliotb-lowrisc: I'd rather not use your derived version for now. This is supposed to be the reference implementation based on my testing. Better for it to have existed in the repo, if only for a brief time, to leave a history of how the test came about.

elliotb-lowrisc:
Tweaked couple of comments

PR #402, opened 2025-04-18
[Legacy] Fix the sram size in linker script

engdoreis:
Fix the sram size in linker script for Legacy mode.

### DIFF
--- a/sw/legacy/test/memory_test.c
+++ b/sw/legacy/test/memory_test.c
@@ -16,9 +16,11 @@
 #include "timer.h"
 
 #define TEST_DATA (0xDEADBEEF)
-#define TEST_SIZE ((224 * 1024) / 4)
+// Total ram size minus space for .text and .rodata
+#define TEST_SIZE_BYTES (0x1F000 - 0x7000)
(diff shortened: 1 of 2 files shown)


sw/legacy/test/memory_test.c
```diff
@@ -16,9 +16,11 @@
 #include "timer.h"
 
 #define TEST_DATA (0xDEADBEEF)
-#define TEST_SIZE ((224 * 1024) / 4)
+// Total ram size minus to space for .text and .rodata
```
alees24, line 19: -to

sw/legacy/test/memory_test.c
```diff
@@ -16,9 +16,11 @@
...
+#define TEST_SIZE_BYTES (0x1F000 - 0x8000)
```
alees24, line 20: This is a little more conservative than necessary; it can at least test another 4KiB, i.e. 1F000 - 7000. I reasoned that as follows - previously the code tested 224 and we've lost 128 so that leaves 224-128=96 that we can test.

alees24 (approved):
Thanks. This is essentially the same as the changes I've made locally. I've left a couple of comments but I'm happy for it be merged when ready.

PR #403, opened 2025-04-21
Enable Malloc on legacy

engdoreis:
Malloc is requited for the heartbleet demo.

### DIFF
--- a/sw/legacy/common/CMakeLists.txt
+++ b/sw/legacy/common/CMakeLists.txt
@@ -4,3 +4,6 @@
 
 add_library(common OBJECT sonata_system.c usbdev.c uart.c timer.c rv_plic.c gpio.c i2c.c pwm.c spi.c crt0.S)
 target_include_directories(common INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}")
+
+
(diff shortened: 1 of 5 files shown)


sw/legacy/common/CMakeLists.txt
```diff
@@ -4,3 +4,6 @@
 
 add_library(common OBJECT sonata_system.c usbdev.c uart.c timer.c rv_plic.c gpio.c i2c.c pwm.c spi.c crt0.S)
 target_include_directories(common INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}")
+
+
+add_library(syscall OBJECT ../../../vendor/cheriot_debug_module/tb/prog/syscalls.c)
```
engdoreis, line 9: I wonder if we should copy this file to the `sw/legacy/common`.

marnovandermaas: Are we likely to make changes to this? If so then maybe it's useful to copy, but if we can leave it unchanged we can keep it in the vendor directory.

marnovandermaas (approved):
Only affects legacy software so happy for this to go in.

PR #404, opened 2025-04-22
PLIC: Remove integrity countermeasure for generate

marnovandermaas:
Previously util/generate_plic.sh would generate an error because the integrity countermeasure was not implemented in the RTL. This removes that countermeasure from the template.

### DIFF
--- a/util/top_gen/generator.py
+++ b/util/top_gen/generator.py
@@ -368,5 +368,7 @@ def generate_top(config: TopConfig) -> None:
             ["clang-format", "sw/cheri/common/platform-pinmux.hh", "-i"],
             check=True,
         )
+        subprocess.run(["sh", "util/generate_plic.sh"], check=True)
+        subprocess.run(["bash", "util/reg_gen.sh"], check=True)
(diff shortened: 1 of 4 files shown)


util/top_gen/generator.py
```diff
@@ -368,5 +368,7 @@ def generate_top(config: TopConfig) -> None:
             ["clang-format", "sw/cheri/common/platform-pinmux.hh", "-i"],
             check=True,
         )
+        subprocess.run(["sh", "util/generate_plic.sh"], check=True)
+        subprocess.run(["bash", "util/reg_gen.sh"], check=True)
```
elliotb-lowrisc, line 372: I'm confused. Is there a reason to use `sh` for one script and `bash` for the other?

elliotb-lowrisc (approved):
Seems to work for me, barring one point of confusion

util/top_gen/generator.py
marnovandermaas: Yes it didn't run under sh. Just copied what it said in the heading of the script.

PR #405, opened 2025-04-24
Pass -nostartfiles when building boot stubs.

resistor:
Per GCC's documentation, when -nodefaultlibs is passed, "The standard startup files are used normally, unless -nostartfiles is used."


marnovandermaas (approved):
Thanks for the fix, approving because CI passes.

resistor:
Thanks! Can you click merge for me?

PR #411, opened 2025-05-05
Run a legacy test to CI

engdoreis:
This will hopefuly prevent the legacy sw stack from breaking.


marnovandermaas (approved):
Thanks, this seems to run successfully. Just a few formatting nits from my end.

PR #412, opened 2025-05-07
[legacy] Add a cpp runtime

engdoreis:
Allows code sharing between CHERIoT and Legacy modes in the Sonata-Software repository.

### DIFF
--- a/sw/legacy/link.ld
+++ b/sw/legacy/link.ld
@@ -57,6 +57,8 @@ SECTIONS
 
         *(.text)
         *(.text.*)
+        /* C++ dependency: keep the static contructors*/
+        KEEP(*(.ctors));
         . = ALIGN(4);
     }  > ram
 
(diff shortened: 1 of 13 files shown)


sw/legacy/link.ld
```diff
@@ -97,6 +99,15 @@ SECTIONS
         _bss_end = .;
         . = ALIGN(4);
     } > ram
+
+    /* C++ dependency: a array with pointers to the static contructors*/
+    .ctors : {
```
nbdd0121, line 104:
```suggestion
    .init_array: {
```

sw/legacy/link.ld
```diff
@@ -97,6 +99,15 @@ SECTIONS
...
+        __init_array_start = .;
+        *(.init_array)
```
nbdd0121, line 106:
```suggestion
        KEEP(*(.init_array))
```

sw/legacy/link.ld
nbdd0121: This whole section should be moved up before .data

marnovandermaas (commented):
I think this looks reasonable to me, just squash the fixup commit.

HU90m (approved):

PR #416, opened 2025-05-27
Replace GPIO synchronisers w/canonical primitive

elliotb-lowrisc:
Something resembling a two stage synchroniser was being used in `gpio_core.sv`, but better to use the matching primitive instead.

### DIFF
--- a/rtl/ip/gpio/rtl/gpio_core.sv
+++ b/rtl/ip/gpio/rtl/gpio_core.sv
@@ -33,7 +33,7 @@ module gpio_core #(
 
   logic [RegAddr-1:0] reg_addr;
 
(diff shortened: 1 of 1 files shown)


elliotb-lowrisc:
Yes, the synchronisers prevent (drastically reduce the probability of) a metastable input signal propagating further into the design, possibly leading to undefined behaviour. The de-bouncering logic relies on a stable (for a given clock edge) input signal to function correctly. See also https://en.wikipedia.org/wiki/Metastability_(electronics)#Synchronous_circuits

marnovandermaas (approved):
Approving because this is a better way of implementing the synchroniser that is already in the design.

PR #417, opened 2025-05-30
Move debounce_step.sv to live in gpio

elliotb-lowrisc:
The debounce_step module is only used in the gpio block, so it should probably live with the gpio files.


marnovandermaas (approved):
Looks good to me.

PR #422, opened 2025-06-09
SPI controller improvements

alees24:
~~This is a draft PR at present for visibility and to solicit feedback, particularly on the architectural changes.~~

The PR makes three changes to the SPI controller:

1. Modifies the implementation of the COPI and CIPO shift registers to simplify the surrounding combinational logic and ensure that the I/O lines each connect to a single flip-flop. This is achieved by performing the bit-reordering on the parallel access to the shift register instead of shifting in both directions.
2. Makes the idle state of the COPI line configurable. The SPI-based access to the microSD card relies upon the COPI line being held in a high state when not transmitting to the card and unfortunately the current SPI controller logic does not offer that functionality. Instead the controller would drop the line which signals the start of a new command and thus interferes with data reception. Receiving data from the microSD card relied upon feeding a continual stream of 0xff-valued bytes into the transmit side whilst reading data from the microSD card, complicating the software and reducing the transfer rate.
~~3. Offers memory-mapped access to the Rx and Tx FIFOs so that the CPU (or other TL-UL agent) may perform (half-)word transfers to/from the SPI controller, offering reduced CPU load and/or higher throughput.~~ *

The final commit updates the associated supporting software to exercise the new functionality.

* This functionality has been dropped whilst it is presently somewhat contentious; there may be a way to introduce the same functionality more compactly by using different TL-UL components or using them differently.

### DIFF
--- a/sw/cheri/checks/sdraw_check.cc
+++ b/sw/cheri/checks/sdraw_check.cc
@@ -59,7 +59,7 @@ extern "C" [[noreturn]] void entry_point(void *rwRoot) {
   WriteUart uart{uart0};
   Log log(uart);
 
-  // The SPI controller talkes to the microSD card in SPI mode.
+  // The SPI controller talks to the microSD card in SPI mode.
   auto spi = spi_ptr(root, 2);
 
   // We need to use the pinmux to select the microSD card for SPI controller 2 reads (CIPO),
--- a/sw/cheri/common/sdcard-utils.hh
+++ b/sw/cheri/common/sdcard-utils.hh
@@ -33,6 +33,8 @@ class SdCard {
   bool crcOn;
   // Access to diagnostic logging.
   Log *log;
+  // Does this SPI controller have the ability to control the idle state of the COPI line?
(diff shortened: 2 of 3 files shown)


elliotb-lowrisc:
Latest utilisation (as I build it) is:

| Slice LUTs | 27,424 | 32,600 | 84.12% | +3.61%pt |
| Slice Registers | 16,155 | 65,200 | 24.78% | +0.88%pt |
| DSPs | 13 | 120 | 10.83% | 0%pt |

Still rather an increase in LUTS.

alees24:
> The "Implement faster FIFO access" commit seems to be the trouble-maker. Building just the first two commits results in a negligible change in utilisation and timing.

As discussed there may be a modified approach that supports faster FIFO access with lower overhead (tlul_socket 1:2 rather than 1:3 and only one tlul_adapter_sram) to reduced the overhead in the TL-UL/prim components. This should hopefully help with both resource usage and timing, but I've dropped it from this PR for now.

PR #424, opened 2025-06-11
[rtl] Increase outstanding requests in SRAM wrapper

elliotb-lowrisc:
Need a minimum of 2 (this is what is used in OpenTitan) to enable back to back requests without stall cycles.

### DIFF
--- a/rtl/system/sram.sv
+++ b/rtl/system/sram.sv
@@ -59,7 +59,8 @@ module sram #(
   // TL-UL device adapters
   tlul_adapter_sram #(
     .SramAw           ( SramAw ),
-    .EnableRspIntgGen ( 0      )
+    .EnableRspIntgGen ( 0      ),
+    .Outstanding      ( 2      )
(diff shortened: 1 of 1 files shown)


alees24:
Background: the rationale behind the contrived code is that capability stores occur as two back-to-back 32-bit word stores. This change has minimal area/resource impact and should benefit interrupt handlers (including therefore thread preemption), as well as regular pointer stores.

alees24 (approved):
I'm happy that this is worth having. Thanks for all the work.

alees24:
> Doubles the performance of capability stores in contrived test code that includes the following loop:

For the record it was later realised that a generic `memcpy` implementation must use clc/csc instructions to transfer 64-bit aligned data because otherwise the capability tag bits will be cleared and memcpy would be incapable of copying correctly any structs/arrays that include pointers. So, this change has a beneficial impact on more sections of code than was earlier realised.

PR #425, opened 2025-06-11
Increase multiplier performance

elliotb-lowrisc:
Switch to the single-cycle multiplier in Ibex by changing the `RV32M` parameter.
Improves multiplication performance while also reducing LUT usage, as much of it can be mapped to the DSP macros in the FPGA.


elliotb-lowrisc:
Current utilisation (as I build it) is:

| Slice LUTs |26,095 | 32,600 | 80.05% | -0.46%pt |
| Slice Registers | 15,599| 65,200 | 23.92% | +0.02%pt |
| DSPs | 16 | 120 | 13.33% | +2.5%pt |

HU90m (approved):

PR #427, opened 2025-06-11
Use BUFGs for some reset signals

elliotb-lowrisc:
Free up some data-routing resource by using global clock-routing resources for some specific high-level reset signals. Not perfect as it drops down to normal data-routing when it gets inverted for use in many places, but does seem to improve timing and routing utilisation a bit.
Might as well, seeing as we were only using 8 of 24 BUFGs.


elliotb-lowrisc:
Current utilisation (as I build it) is:

| Slice LUTs |26,210 | 32,600 | 80.40% | -0.11%pt |
| Slice Registers | 15,584 | 65,200 | 23.90% | 0%pt |
| DSPs | 13 | 120 | 10.83% | 0%pt |

elliotb-lowrisc:
The questionable area savings are probably not worth the extra code complexity

PR #428, opened 2025-06-22
Reduce the cost of TL-UL interface to device registers

alees24:
A number of the devices in the Sonata System have been inherited from a system using a simple RAM-like interface for its bus. The interfacing to TL-UL is a little redundant for the following devices:

- PWM
- GPIO
- rv_timer

The `tlul_adapter_reg` always drives the `d` channel valid/data a single cycle after acceptance of the transaction and thus has its own internal register stage, which is also required to hold the data stable until it is accepted by the TL-UL fabric. This means that the devices themselves are no longer required to have the additional register delay, nor the associated combinational muxing within the `tlul_adapter_reg`.

Switch to `AccessLatency` 0 instead, and simplify the device logic for PWM and rv_timer. _GPIO is left until later_.

### DIFF
--- a/rtl/system/rv_timer.sv
+++ b/rtl/system/rv_timer.sv
@@ -10,7 +10,9 @@ module rv_timer #(
   // Bus data width (must be 32)
   parameter int unsigned DataWidth    = 32,
   // Bus address width
-  parameter int unsigned AddressWidth = 32
+  parameter int unsigned AddressWidth = 32,
+  // Number of cycles by which the read data lags the request (must be 0 or 1)
+  parameter int unsigned AccessLatency = 1
 ) (
   input  logic                    clk_i,
   input  logic                    rst_ni,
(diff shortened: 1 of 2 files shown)


rtl/ip/pwm/rtl/pwm_wrapper.sv
```diff
@@ -0,0 +1,109 @@
+// Copyright lowRISC contributors.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+// This wrapper instantiates a series of PWMs and distributes requests from the device bus.
+module pwm_wrapper #(
+  parameter int unsigned PwmWidth   = 12,
+  parameter int unsigned PwmCtrSize = 8,
+  parameter int unsigned DataWidth  = 32,
+  parameter int unsigned RegAddrWidth = 12
```
elliotb-lowrisc, line 10: Could `RegAddrWidth` be derived from `PwmWidth`?

PR #432, opened 2025-07-09
Reduce access latency of GPIO blocks

elliotb-lowrisc:
Rework the `gpio_core` to provide response data in the same cycle as the request to reduce the access latency.

Reduce the value of the `AccessLatency` parameter of the TL-UL adapter used to connect the GPIO blocks to the system bus, to match the `gpio_core` read latency changes.

### DIFF
--- a/rtl/ip/gpio/rtl/gpio_core.sv
+++ b/rtl/ip/gpio/rtl/gpio_core.sv
@@ -18,7 +18,6 @@ module gpio_core #(
   input  logic                 device_we_i,
   input  logic [3:0]           device_be_i,
   input  logic [DataWidth-1:0] device_wdata_i,
-  output logic                 device_rvalid_o,
   output logic [DataWidth-1:0] device_rdata_o,
 
   input  logic [GpiWidth-1:0] gp_i,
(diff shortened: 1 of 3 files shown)


elliotb-lowrisc:
FPGA bitstream build indicated a possible minor increase in Slice LUTs from 80.57% to 80.90% (+0.33%pt), but this is small enough that it may in part be build noise. Timing looked fine.

rtl/ip/gpio/rtl/gpio_core.sv
```diff
@@ -138,12 +136,12 @@ module gpio_core #(
...
         pcint_status <= pcint_status;
       end
-
-      device_rvalid_o <= device_req_i && !device_we_i;
-      rd_reg_idx <= reg_addr[RegAddr-1:2];
     end
   end
 
+  assign device_rvalid_o = device_req_i && !device_we_i;
+  assign rd_reg_idx = reg_addr[RegAddr-1:2];
```
alees24, line 143: These signals are no longer required.
`device_rvalid_o` has never been used in the parent module, and `tlul_adapter_reg` has no rvalid input, but rather assumes the timing of the read data availability.
`rd_reg_idx` is used internally in an if/else ladder at present.

Could I ask, please, that - as discussed previously - that be simplified to a `case` statement, which may be controlled directly by `reg_addr`? Also, the testing of `device_rvalid_o` at the top of that always_comb block can be dropped which I think will save a little logic too.

alees24 (approved):
This looks functionally fine to me, but there are a couple of things that could be tidied, please, whilst it's being changed.

PR #435, opened 2025-07-11
Push the simulator build result to cache

nbdd0121:
This uses lowRISC's public nix cache to provide caching for simulator and simulator tests.

To do this, however, we need to stop the derivation from depending on the git hash, otherwise every commit will trigger a rebuild due to hash change. I think this should be fine, as the system info blocks need a git commit hash for release propose, but we don't need this for simulator?


elliotb-lowrisc (approved):
Let's give it a go!

machshev (approved):

PR #438, opened 2025-07-25
Corrections to interrupt handling in plic_tests

alees24:
Fix the disabling of interrupts in the `_interrupt_set` functions; these were previously incapable of clearing the mie|mstatus bit. Modify the interrupt timeout loop in plic_tests.hh accordingly.

Re-express mcycle 'wfiTimeout' check to be robust against counter wraparound (this takes only about 100 seconds of run time on the Sonata board at 40MHz).


AlexJones0 (approved):
Thanks for fixing both these bugs! The wait avoiding the counter wraparound is also much nicer now.

PR #441, opened 2025-07-30
Linear execution test of HyperRAM interface

alees24:
Add a new test that both checks and measures the performance of executing a linear sequence of repeated instructions from the HyperRAM.

This is perhaps more relevant for most current uses of the HyperRAM because it is largely intended to be used for additional code rather than loading/storing data.

### DIFF
--- a/sw/cheri/checks/hyperram_test.cc
+++ b/sw/cheri/checks/hyperram_test.cc
@@ -28,6 +28,21 @@ using namespace CHERI;
 const int RandTestBlockSize = 256;
 const int HyperramSize      = (1024 * 1024) / 4;
 
+// Ensure that all writing of code to memory has completed before commencing execution
+// of that code. Code has been written to [start, end] with both addresses being
+// inclusive.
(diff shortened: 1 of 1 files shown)


sw/cheri/checks/hyperram_test.cc
```diff
@@ -739,6 +794,25 @@ extern "C" [[noreturn]] void entry_point(void *rwRoot) {
...
+      cache_enabled = !cache_enabled;
+      icache_enabled_set(cache_enabled);
+      icache_invalidate();
+      log.println("Running linear execution test with icache {:s}...", cache_enabled ? "enabled" : "disabled");
+      failures += linear_execution_test(hyperram_area, prng, log, true, 0u, lin_exec_len, lin_exec_iters);
+      log.print("  result...");
+      write_test_result(log, failures);
+    } while (cache_enabled);
+    // Reinstate the normal icache operation.
+    icache_enabled_set(true);
```
marnovandermaas, line 814: Should we also icache_invalidate here?

marnovandermaas (approved):
One question but otherwise looks good to me.

PR #447, opened 2025-08-25
CLA added

marnovandermaas:


elliotb-lowrisc (changes requested):
Two things:

First, there should probably be some mention of the CLA and how to indicate your agreement with it, similar to CONTRIBUTING.md in OpenTitan.

Secondly, if people are going to have to agree to this to submit work, then we should probably add a CI check for that.

PR #460, opened 2026-03-16
[DPI] Create .core file for i2cdpi to help reuse

elliotb-lowrisc:
Having a .core file specifically for the I^2C DPI module makes it easier to use in other projects.


marnovandermaas (approved):
Looks good to me. Are you able to test the simulation target manually before merging?

